PR #57, opened 2026-06-11
Restore the global-footprint world map to the homepage

jperla:
## Why

The 2026-06 homepage redesign (`14bd9ba`) dropped the world-map section. It's a key scale/credibility signal — it visually proves the multi-region footprint the page's reliability section claims ("Live regions: N"). Bringing it back.

## What

The redesign removed only the **markup** — everything that feeds the map survived:
- `.regions-map*` CSS (`dashboard.css`, 29 selectors, untouched)
- `data-regions-stage` interaction JS (`dashboard.js`, click-to-highlight)
- `world-map.svg` asset
- `map_regions` context (`dashboard.py:455`)

So this re-inserts the original markup verbatim (data-driven from `map_regions` — region count and marker coordinates stay correct), wrapped in the redesign's `home-section`/`home-wrap` so spacing/width match the new page. Placed right after the "Fallback when providers fail" reliability section.

## Also fixes a red smoke test

`test_marketing_page_advertises_production_not_alpha` asserts the homepage contains `world-map.svg`/`<svg` and has been failing since the redesign stripped the map. This re-greens it.

## Verification

Rendered `dashboard_html()` offline and screenshotted the isolated section:
- "Global footprint: N regions" heading
- 7 markers positioned **inside** the SVG (us-central1 = primary pulse; others serving/edge in correct geographic spots)
- Interactive region list with primary/live/edge pills alongside
- No Jinja errors; the marketing-smoke assertion now passes against the rendered output

map renders correctly

### DIFF
--- a/src/trusted_router/templates/dashboard.html
+++ b/src/trusted_router/templates/dashboard.html
@@ -236,6 +236,44 @@ const response = await client.chat.completions.create({
       </div>
     </section>
 
+    {# 6b. Global footprint map — visual proof of the multi-region scale
+       claimed in the reliability section above. Markup restored after the
+       2026-06 homepage redesign dropped it; the .regions-map CSS, the
+       data-regions-stage JS in dashboard.js, world-map.svg, and the
+       map_regions context (dashboard.py) were all left intact. #}
+    <section class="home-section">
+      <div class="home-wrap">
+        <section class="regions-map">
+          <div class="regions-map-head">
+            <h2>Global footprint: {{ map_regions|length }} regions</h2>
+            <p>Verified live gateways plus regional expansion points for lower latency and faster failover. Public status separates router health from provider health.</p>
+          </div>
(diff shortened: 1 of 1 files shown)


jperla:
Superseded — the world map is already back, in a better form than this PR restores.

main's `src/trusted_router/static/hero-router-scene.js:226` renders a *"Globe continents overlay"* and loads `/static/world-map.svg` at line 270, so the multi-region footprint this PR wanted to make visible is now the animated globe in the hero rather than a static section further down the page. The live homepage confirms it: the hero is `hero-copy`/`hero-badges`/`hero-line`, with no separate world-map section.

Re-adding the markup from the pre-2026-06 layout would put a second, static copy of the same signal on the page. The asset this PR depended on survived the redesign and is in active use; only the placement changed. Closing as superseded rather than stale — the underlying point was right, and it got addressed.

PR #73, opened 2026-06-21
pricing: refresh catalog to current (2-week baseline) + carry forward servable models

jperla:
First catalog refresh since **2026-06-07** (the freeze fixed in #72), run through the fixed pipeline with all provider keys and **human-reviewed** (signed off).

## Catalog: 99 → 138 models
- **+41** current releases the frozen catalog was missing: `claude-opus-4.8`, `gpt-5 / 5.1 / 5.2 / 5.3-codex`, `gemini-3-pro-image`, `glm-5.2`, qwen3 family, and more.
- **−2** only: `arcee-ai/trinity-large-thinking`, `stepfun/step-3.5-flash` — genuinely dead (no provider-direct price; can't bill).
- **Preserved** (would've been dropped by OR-delisting): `claude-opus-4`, `claude-sonnet-4`, and the free tier (`z-ai/glm-4.5-air:free`).

## Durability fixes (so refreshes stop shedding servable models)
- **Carry-forward**: when OR delists a model a provider still prices, carry the previous snapshot's metadata forward and reprice from the live provider result instead of dropping it.
- **Free tier**: `:free` SKUs are the one legitimate `$0` — exempt from the `$0`-unpriced guard so `trustedrouter/free` keeps candidates.
- **Tie-break**: `cheapest` now breaks prompt ties on completion, killing phantom completion spikes (gemma-3-27b-it) that would re-trip the watchdog.

## One real price move (reviewed)
`qwen2.5-vl-72b` prompt **4×** — phala dropped the model (verified against its live catalog), parasail is the floor now. Correct, not a bug.

## Tests
- Together tests rewritten to assert the **wiring** (Credits/BYOK + positive provider-direct price) against whatever Together serves now — not a pinned model/price that breaks every refresh (the old ones pinned llama-3.3, which Together dropped).
- New carry-forward + free-tier regression tests.
- Full suite green (1002 passed), ruff + mypy clean.

## Deploys on merge
This snapshot is under `src/**`, so merging triggers the staged-canary control-plane deploy. The catalog jump is intentional and reviewed.

### DIFF
--- a/scripts/pricing/refresh.py
+++ b/scripts/pricing/refresh.py
@@ -522,6 +522,7 @@ def _merge_snapshot(
     or_snapshot: dict[str, Any],
     provider_index: dict[str, dict[str, ModelPrice]],
     healed_slugs: set[str],
+    prev_snapshot: dict[str, Any] | None = None,
 ) -> dict[str, Any]:
     """Build the final snapshot.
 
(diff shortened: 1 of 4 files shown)


jperla:
Closing as obsoleted by automation. This is a hand-built pricing baseline from 2026-06-21 — 10,822 insertions, 1,073 commits behind main. The snapshot it rebuilds is now refreshed hourly by the pricing bot (main's latest: `[bot] hourly price refresh: 2026-08-24T11:23Z`). Rebasing would mean reconciling a two-month-old snapshot against a file rewritten several hundred times since. If any of the carry-forward logic in `scripts/pricing/refresh.py` is still wanted, it is worth re-cutting against today's refresh rather than salvaging this.

PR #74, opened 2026-06-23
[codex] Add x402 stablecoin funding

jperla:
## Summary

Adds x402 stablecoin funding as a control-plane-only payment rail for existing TrustedRouter API-key users.

- Adds `/v1/billing/x402/fund` to create a Stripe crypto PaymentIntent and return an x402 `payment-required` challenge.
- Adds `/v1/billing/x402/settle` to verify Stripe settlement and credit the API key workspace exactly once.
- Wires Stripe webhook handling for x402 PaymentIntent success, pending/failure statuses, orphan events, and refund manual-review visibility.
- Adds fail-closed configuration for production and local/test-only mock payment guards.
- Documents the design and adds a public `/docs/x402` page.
- Adds focused tests for workspace isolation, idempotency, wrong asset/network/currency, malformed IDs, rate limits, and prompt/output non-leakage.

## Boundary

x402 remains outside the enclave. The attested API plane still only sees normal credit availability and does not parse payment headers or send prompt/output content to Stripe.

## Validation

- `uv run ruff check .`
- `uv run mypy`
- `uv run pytest tests/test_billing.py tests/test_x402_billing.py -q`
- Earlier full-suite run before branching: `1019 passed, 33 skipped`

## Notes

Two unrelated local edits were intentionally left unstaged: the blog draft and `scripts/credit_grant_joseph.py`. Claude CLI review was attempted after local validation, but the first invocation had a bad stdin expansion, the second exceeded a low budget, and the final higher-budget invocation hung without findings before being stopped.

### DIFF
--- a/docs/design/x402-stablecoin-funding.md
+++ b/docs/design/x402-stablecoin-funding.md
@@ -1,0 +1,297 @@
+# TrustedRouter x402 Stablecoin Funding Design
+
+Status: draft for review
+Owner: Lore Hex Corp
+Last updated: 2026-06-22
+Review status: reviewed with `claude --print` on 2026-06-22; P0/P1 feedback incorporated.
+
+## Summary
+
+TrustedRouter should support x402 stablecoin payments as a control-plane funding rail for existing workspaces. x402 should not run inside the attested enclave and should not change the prompt path. Agents still use a normal TrustedRouter API key. When credits are low, the agent can call a separate billing API to receive a Stripe x402 payment challenge, pay in USDC, and have the workspace credited after Stripe verifies settlement.
+
+The attested gateway remains unchanged: it receives the API key, asks the control plane for metadata-only authorization, serves the prompt only when credits are available, and returns insufficient-credit errors when the ledger cannot reserve funds.
+
+## Goals
+
+- Let agents automatically add prepaid credits using stablecoin.
+- Require a TrustedRouter API key for all x402 flows.
+- Keep the existing integer credit ledger as the source of truth.
+- Keep all x402 logic in the control plane, outside the enclave.
+- Keep prompt and output content out of Stripe, x402 payment metadata, Sentry, logs, and billing records.
+- Make all settlement paths idempotent.
+- Credit only from Stripe-confirmed settled amounts, never from client input.
+
+## Non-Goals
+
+- No accountless anonymous paid inference in v1.
+- No per-token on-chain refund path in v1.
+- No x402 code inside the enclave.
+- No prompt-path fallback to a non-attested control-plane route.
+- No Coinbase facilitator in v1. Stripe x402 is first.
+- No sub-cent x402 top-ups in v1, because Stripe PaymentIntent amounts are cent-denominated.
+
+## Research Notes
+
+- x402 uses HTTP `402 Payment Required` with machine-readable payment requirements so a client or agent can pay and retry. Coinbase describes this as a buyer requesting a resource, the server returning `PAYMENT-REQUIRED`, then the buyer sending payment authorization via `PAYMENT-SIGNATURE`.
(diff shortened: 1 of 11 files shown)


jperla:
## Claude CLI review

**Verdict: Approve with nits.** This cleanly respects the control-plane / enclave boundary — no x402 logic touches the prompt path, the new endpoints live in `routes/billing.py` + `services/x402_billing.py`, and the enclave's authorize-then-serve contract is unchanged.

**Boundary & data-flow review (looks correct):**
- API key required on both `/billing/x402/fund` and `/billing/x402/settle` (`routes/billing.py:109,141` — explicit `principal.api_key is None → 403`).
- Workspace is derived from the key; `X402FundingRequest(_Strict)` (`schemas.py`) rejects body-supplied `workspace_id` before any Stripe call. Test `test_x402_fund_rejects_workspace_id_and_cent_fraction_before_stripe` covers this and confirms zero Stripe calls.
- Stripe metadata is limited to `workspace_id / amount_microdollars / payment_method / purpose / asset / network` (`services/x402_billing.py::_create_payment_intent`). No prompts, no raw API keys, no key hashes. Asserted by `assert "secret prompt" not in json.dumps(captured)` / `"sk-tr-"` in the fund test.
- Crediting uses `STORE.credit_workspace_once(..., x402_event_id(pi_id))` — same idempotency key for webhook and `/settle`, so both-orders race and 4-way concurrent settle both credit once (tests cover both).
- Credited amount = `min(amount_received or amount, requested_microdollars_from_metadata)`; client `/settle` body cannot influence the credited amount. Status gate (`!= "succeeded" → credited=false`) is enforced before the min().
- Settle returns `404` (not `403`) for cross-workspace PI references via `hide_cross_workspace=True`, so it doesn't leak PI existence to another tenant.
- `payment_intent_id` regex `^pi_[A-Za-z0-9_:-]{1,253}$` blocks Stripe call on malformed input.
- Fail-closed config: `x402_allow_mock_payments` is rejected outside `local|test`; `x402_enabled` without Stripe creds is rejected outside `local|test` (`config.py::production_is_fail_closed`).
- Refund / orphan paths return 200 with operator-visible log + flag (`x402.refund_requires_manual_review`, `x402.orphan_payment_intent`) — Stripe won't retry, but no silent debit/credit.

**Nits (non-blocking):**
- P2 — `services/x402_billing.py::workspace_id_from_mock_payment_intent` derives a workspace id from the mock PI suffix. The comment claims it "deliberately cannot target a real workspace without a monkeypatch," which is true in practice only because (a) `_mock_payments_enabled` is gated to `local|test` and (b) `/settle` always passes `expected_workspace_id` from the caller's API key. Worth a one-line assert or comment tying those two guards together so a future refactor doesn't drop the workspace-match check and turn this into a "credit any workspace in mock mode" footgun.
- P2 — `schemas.py::X402FundingRequest.amount_in_range` validates against `MAX_CHECKOUT_DOLLARS`, and `validate_x402_funding_amount` re-validates against `x402_max_fund_dollars`. The double-check is fine (schema = hard ceiling, settings = workspace cap), but the duplicated cent-exactness check could be dropped from one side; current state is harmlessly belt-and-suspenders.
- P2 — `webhook.py` catches `HTTPException` on the orphan path and inspects `status_code != 404`. Works today because `api_error(404, ...)` is the only 404 raised in `credit_x402_payment_intent`, but if a future change starts raising 404 for a different reason (e.g., a missing dependent record) the webhook would also mark it `orphan=True`. Consider a more specific sentinel or matching on `ErrorType.NOT_FOUND`.

No blocking findings. The integer-ledger-as-source-of-truth invariant, the "credit only Stripe-settled amount capped by requested" rule, and the "prompts never touch Stripe/logs/docs" rule all hold in this diff.

PR #85, opened 2026-06-25
billing(step6): refuse key deletion while a typed hold is in flight

jperla:
## Why
Deleting an API key drops its `tr_key_limit` row (`mirror_delete`). If a typed hold is still in flight, the settle's key release matches **0 rows** (`release row-count != 1`) and strands the hold — the same incident class, key-side. codex flagged this as a hard blocker before broad rollout.

## Fix
`delete_key` refuses (**503 + Retry-After=5**) when `STORE.key_has_open_typed_hold(hash)` is true (an open `tr_reservation`, `settled=false`); the client retries after it drains (seconds). New store method on all three impls (GCP queries `tr_reservation`; InMemory has no typed tables → `False`; Protocol). The `/v1/keys/{hash}` DELETE is the single delete chokepoint.

## Tests
Unit (open hold → True; settled/none → False); route refusal (503 + Retry-After, key not deleted); route success when no hold. Lint clean.

Last focused safety blocker of the Step-6 series (after #79–#84). Whales still need #33 (fast kill-switch gate + settle outbox); the ramp is a separate metered operation.


jperla:
Closing unmerged. codex review found the disable-first mitigation still has a TOCTOU residual that leaves an **unreapable** typed key-hold orphan (a `release_key` on the mirror-deleted `tr_key_limit` row returns row-count 0 → `SettleOutcome.ERROR`; the reaper only reclaims on `SETTLED`, so it strands until manual repair). The correct fix is an **atomic count+delete in one read-write transaction** (the count's read-lock serializes against a concurrent authorize's `tr_reservation` insert) — an IO-layer change best done with fresh context. Not a first-low-traffic-batch blocker (workspaces are quiesced during a flip, so no in-flight holds); tracked for the whale-rollout hardening. The disable-first work + tests remain on branch `billing-key-delete-drain` for that follow-up.

PR #89, opened 2026-06-26
billing(flip): reconcile_typed_credit_usage — post-flip usage catch-up

jperla:
The post-sweep fix for residual typed usage drift after the universal flip. Sets typed credit total_usage = JSON.total_usage + SUM(settled-Credits actual_micro) for a billing-PAUSED workspace (shard 0). Fixes the 2 ever-typed leftovers ($3.98) and any new-workspace straddle from the rollout window. Mirrors repair_typed_reserved's fail-closed shape (in-txn re-read, abort on missing typed row / nonzero-shard / not-paused); leaves reserved + keys untouched. CLI: `usage WS --apply`.

### DIFF
--- a/docs/design/billing-typed-cutover-handoff.md
+++ b/docs/design/billing-typed-cutover-handoff.md
@@ -1,0 +1,385 @@
+# Typed-billing cutover handoff — resolved synthetic cohort
+
+**Status as of 2026-06-21 10:55 Pacific.**  
+Owner handing back: Codex. Next reader: Claude.
+
+## TL;DR
+
+The typed-column conditional-DML billing path is now active in production for the
+synthetic monitor workspace:
+
+`45819281-0ce9-4811-a0cd-c660ab3a116d`
+
+The earlier "typed authorize path is deployed but not engaging" symptom is
+resolved. The real cause was not a bad cohort flag or stale image. The cohort
+runtime flags were correct, but the synthetic workspace's **typed credit mirror**
+was stale/exhausted. Typed authorize correctly read the typed counters, found no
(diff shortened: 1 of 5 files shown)


jperla:
Closing as stale — but there is a real gap here that should outlive the PR, so recording it rather than just dropping it.

**Why this can't be merged as-is.** It was a one-shot catch-up for the June typed-billing flip: the body targets "the 2 ever-typed leftovers ($3.98)" and a straddle from that rollout window. That moment has passed, `ever_typed` no longer appears anywhere in `src/`, and the 385-line handoff doc describes a cutover that has since completed. The code has also drifted badly: this PR *creates* `src/trusted_router/storage_gcp_counter_reconcile.py` with 119 lines, while main independently grew that same file to 382 lines with a different design (`audit_typed_invariants`, `repair_typed_reserved`). A rebase would be a rewrite, not a merge.

**The gap that is still real.** main's `audit_typed_invariants` asserts only that every typed `reserved` equals its open holds, in both directions — it never audits `total_usage`. `repair_typed_reserved` likewise repairs reserved only. So the repo currently has drift detection and repair for *reserved* counters and none for *usage* counters, which is exactly what this PR's `reconcile_typed_credit_usage(store, workspace_id, *, apply=False)` was for.

If usage drift is worth guarding, the way to do it now is a small function alongside `repair_typed_reserved` in the module as it exists today, plus an audit arm in `audit_typed_invariants` — not this branch. Reopen or re-cut against current main if you want that; the design reasoning in the handoff doc is still the best write-up of why the invariant holds.

PR #107, opened 2026-07-03
Extract catalog_registry.py: isolate MODELS/MODEL_ENDPOINTS build (#38)

jperla:
## What

Splits the money-critical registry construction out of the `catalog.py` god-module (1592 → ~1000 lines).

`catalog_registry.py` now holds the ~575-line import-time build of `MODELS` / `MODEL_ENDPOINTS` (the hand-coded Auto seed + snapshot/supplemental ingest + uniform cost×1.10/$0.01-floor pricing). `catalog.py` re-exports both and keeps the privacy / routing / serialization query helpers on top.

## Why it's safe

- **Byte-identical move** of the build region (extracted programmatically, not retyped).
- The construction uses **zero `catalog.py` functions** (verified by AST scan), so the registry depends only on the `catalog_data` / `catalog_ingest` / `pricing` leaves — **no import cycle** (a new test pins this).
- `catalog.MODELS is catalog_registry.MODELS` — single source of truth, re-export not copy.

## Verification

- Full suite: **1190 passed, 33 skipped** (identical to pre-split — behavior-preserving).
- ruff + mypy clean.
- New `tests/test_catalog_registry_split.py`: standalone-import (no cycle), re-export identity, query-helpers-stay-out guardrail.

## Follow-up

Sets up the routing extraction (#38's named target): routing reads `MODELS`, which now has a cycle-free home to import from.


jperla:
**codex review: money-path PASS; back-compat FAIL waived (verified non-issue).**

codex confirmed the important guarantees:
- Construction body **byte-identical** from `# Catalog seed` through `MODEL_ENDPOINTS = _filter_unserved_provider_endpoints(...)`.
- **No import cycle** — registry imports only `catalog_data` / `catalog_ingest` / `pricing`; importing the registry standalone does not load `catalog`.
- `catalog.MODELS is catalog_registry.MODELS` and same for `MODEL_ENDPOINTS` (single source of truth).

The one FAIL: my PR text over-claimed "nothing that used to be importable is now missing." codex correctly noted a handful of **private** build artifacts (`_INGESTED_MODELS`, `_INGESTED_ENDPOINTS`, `_SUPPLEMENTAL_MODELS`, `_SUPPLEMENTAL_ENDPOINTS`, `_EMBEDDING_MODELS`, and the leaked loop bindings `_model` / `_model_id`) are no longer attributes of `trusted_router.catalog`. codex itself flagged this as waivable "if private construction artifacts are intentionally outside the compatibility contract."

They are. Verified:
- `grep` across `src/ tests/ scripts/`: **nothing** references any of these via `catalog.*` or imports them from `catalog` — they're `_`-private construction internals.
- The **public** contract is 100% intact: `dir(catalog)` non-underscore names = **124 on main, 124 on this branch, zero lost**.
- `_model` / `_model_id` are accidental module-scope loop-binding leaks; re-exporting them would be wrong (they were never API). Left in the registry as-is to preserve the byte-identical move.

Scope of the back-compat guarantee is therefore the **public** surface, which holds. Merging.

PR #113, opened 2026-07-03
Durable settle outbox — Increment 1: native-table storage (dormant) (#44)

jperla:
First slice of the adversarially-hardened v2 design ([docs/design/durable-settle-outbox.md](docs/design/durable-settle-outbox.md)). **Purely additive and dormant** — nothing calls the outbox on the live settle/reaper path yet, so it's dead code in prod until later increments wire it (and even then only when `settle_outbox_enabled` flips). The prod-enable flip stays owner-gated.

## What's here
- **DDL** — native `tr_settle_outbox` table, PK `(authorization_id, intent_kind)` so a settle and a refund never clobber (MF7 + SF1), plus a `(status, next_attempt_at)` due-scan index (SF10). Additive/guarded in `migrate_typed_counters.sh`.
- **config** — `settle_outbox_enabled` (default `False`).
- **`SettleOutboxRow`** — carries the **frozen** settle inputs (`actual_cost_micro` / `settle_origin` / `reservation_id` / `selected_endpoint_id` / `model_id` / `selected_usage_type`) captured at enqueue (MF4/MF5).
- **`SpannerSettleOutbox`** — INSERT-as-claim enqueue that raises `ALREADY_EXISTS` and refreshes a still-pending row's frozen inputs (MF7, SF9); lease-fenced conditional-DML `claim`/`mark` with exponential backoff → `dead` at max_attempts; `has_intent()` = the reaper-guard predicate (freezes on `pending`|`dead`, **not** `done`|`release_approved`, MF3).
- **Fake Spanner models the table explicitly** (INSERT/UPDATE/SELECT + the guard predicate) so a status/column/guard typo **fails a test** rather than silently matching a generic branch (MF6). The column list is imported from the storage module to prevent schema drift.

## Verification
- 9 storage-layer tests: enqueue idempotency + refresh + terminal no-clobber, settle/refund polarity, due/claim lease, mark done/backoff/dead, lost-lease reject, `has_intent` freeze semantics.
- Full suite: **1210 passed, 33 skipped**. ruff + mypy clean.

## Next increments (separate PRs, each codex-gated)
1. Reaper in-txn guard (MF1/MF2/MF3) — the correctness spine.
2. Frozen-cost finalize primitive + richer outcome (MF5/SF3/SF7).
3. Enqueue-at-settle + drain endpoint (behind the flag).

Will not merge until codex reviews this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #116, opened 2026-07-04
Settle outbox Increment 2: reaper in-txn guard (MF1/MF2/MF3)

jperla:
Increment 2 of `docs/design/durable-settle-outbox.md` (§4 — read §3 for why this guard is the SOLE lost-charge correctness mechanism). Increment 1 (native-table storage, #113) is merged; this wires the reaper interlock. Authored by codex-cli against a decision-final spec; adversarially reviewed by Claude (one fail-closed fix applied, see below).

## What
- `settle_atomic(guard_outbox=False)` — when True (reaper only), a strong read of `GUARD_COUNT_SQL` **inside the read-write claim transaction** aborts the free-release with `SettleOutcome.OUTBOX_GUARDED` if the authorization has a `pending`/`dead` outbox row. In-txn is the MF2 interlock: Spanner serializes it against a concurrent enqueue commit that a snapshot pre-filter would miss. Every live settle/refund caller keeps `guard_outbox=False` — the **inline** settle must proceed (it's what resolves the intent); only the reaper's zero-charge release is guarded.
- `reap_expired_reservations` now projects `authorization_id` (the design calls out it only projected `reservation_id`), advisory-skips candidates with live intents (latency only), and reaps with `guard_outbox=True`.
- **Rollout-safe probe**: the DDL is operator-applied and may lag the deploy. A per-reap probe arms the guard only when the table exists; table-missing (message must *name* `tr_settle_outbox`) falls back to today's exact behavior — correct, since no intent rows can exist without the table.
- **Review finding fixed (fail-closed)**: transient Spanner errors are also typed `NotFound` (e.g. "Session not found"); classifying those as table-missing would have silently disabled the interlock for the cycle. Now any probe error not naming the table **re-raises** — a transient can delay reaping, never unguard it.
- `GUARD_COUNT_SQL` is single-source (has_intent + advisory + in-txn all execute the same literal); the fake `_require_pred`s both predicates and the new scan projection (MF6: a dropped predicate fails a test, it is never silently absorbed).

## Inertness
Byte-identical when: `guard_outbox=False` (all live callers), the table is empty (`NOT EXISTS` vacuous), or the table doesn't exist yet. Enqueue ships default-off in Increment 4; nothing populates the table today.

## Tests
9 new (`tests/test_settle_outbox_guard.py`): pending/dead freeze (twice — frozen, not deferred), `release_approved`/`done` reap, empty-table baseline, **MF2 stale-advisory race** (advisory stubbed False → in-txn check still blocks), `guard_outbox` semantics both ways, table-missing fallback, transient-NotFound fail-closed. Full suite **1228 passed**, ruff + mypy clean repo-wide.

Per the handoff hard constraints: codex review will be posted on this PR before merge; no auto-merge armed.

### DIFF
--- a/src/trusted_router/storage_gcp_authorize.py
+++ b/src/trusted_router/storage_gcp_authorize.py
@@ -326,32 +326,39 @@ def settle_atomic(
 
 def _is_table_missing(exc: Exception) -> bool:
     # Rollout guard: code can deploy before the operator-applied outbox DDL.
-    # FAIL CLOSED on anything else (e.g. a transient "Session not found" is
-    # also typed NotFound): only a message naming the table counts — real
-    # Spanner raises NotFound("Table not found: tr_settle_outbox"). Any other
-    # probe error re-raises, so a transient can only delay reaping, never
-    # silently disable the interlock.
-    text = str(exc)
-    lowered = text.lower()
-    if "tr_settle_outbox" not in text:
-        return False
+    # FAIL CLOSED on everything except the ONE real "table itself is missing"
+    # shape: Cloud Spanner (and its emulator) raise NotFound("Table not found:
+    # tr_settle_outbox") — table name AFTER the phrase. Anchoring on position
+    # keeps wrapped transients ("Session not found while querying
+    # tr_settle_outbox") and schema errors ('column "status" of relation
+    # "tr_settle_outbox" does not exist') from silently unguarding a cycle;
+    # any other probe error re-raises, so a mismatch can only delay reaping,
+    # never free-release a guarded hold.
+    lowered = str(exc).lower()
     return (
-        type(exc).__name__ == "NotFound"
-        or "not found" in lowered
-        or "does not exist" in lowered
+        "table not found" in lowered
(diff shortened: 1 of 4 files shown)


jperla:
codex review: PASS (2 rounds)

- Round 1: FAIL — P1 (`_is_table_missing` failed open for wrapped transients naming the table), P2 (guarded rows could exhaust the reaper LIMIT and starve unguarded expired holds).
- Round 2 (on a1ed031): PASS — "No remaining correctness, safety, or liveness issues were found. The fail-closed probe and starvation-free guarded scan align with the design invariants, and the full test suite passed." Codex independently re-ran pytest (1231 passed) and ruff.

Additional adversarial verification (3-lens agent pass) drove the round-3 tightening included in a1ed031: probe anchored to the single real Spanner table-missing shape, MF6 predicate anchors for `expires_at < @now` / `LIMIT @limit` in the fake, fired-flag assertion in the MF2 stale-advisory test, and insertion-order-robust starvation coverage.

PR #118, opened 2026-07-04
Durable settle outbox — Increment 3: frozen-cost finalize primitive (dormant)

jperla:
Increment 3 of docs/design/durable-settle-outbox.md (§3, §6, §11). Increments 1 (#113) and 2 (#116) are merged. This adds the narrow, DORMANT apply primitive the Increment-4 drain will call — no live caller beyond a behavior-identical refactor in gateway.py.

**What's in it**
- `services/settle_outbox_apply.py`: `apply_frozen_settle(row)` — applies the frozen `actual_cost_micro` via `typed_finalize_gateway` (rich outcomes, never the bool-collapsing wrapper). Full 8-value `ApplyOutcome` contract documented in design §6.
- Typed lost-claim disambiguation via `read_typed_reservation` point-read (new store method + protocol entries): charged replay vs reaper free-release, with the zero-cost benign rule (a frozen \$0 row replay is never a false invariant alert).
- `already_settled_legacy`: legacy origin honestly reports that charged-vs-free is indistinguishable (legacy free releases exist via inline refund/failure-settle; legacy reservations record no amount).
- MF4 park on capability-missing AND transient typed-store outages (6-exception api_core tuple incl. ResourceExhausted and RetryError) at pre-read, finalize, and disambiguation read — typed rows never burn attempts toward dead during an outage, never reroute to legacy.
- MF5 frozen generation rebuild: catalog-independent, retired endpoints string-parse the slug, validated-body parity with inline, deterministic-bad bodies → `invalid_row`.
- SF7: no pricing/auto-refill/budget-alert/broadcast imports; sole side effect is the claim-gated once-only activity index + benchmark sample on `settled_now` (inline mirror).
- Extraction: gateway.py cache-token normalization → shared `normalized_prompt_accounting()` (behavior-identical; full suite + a lenient-coercion pin test prove it).

**Verification**
- Repo-wide gates green: ruff, mypy (161 files), pytest 1255 passed / 33 skipped.
- Three review rounds: my line-by-line review plus two adversarial multi-agent verification passes (5 verifier lenses total); all findings fixed (legacy mislabel, availability-gated park, zero-cost conflation, generation parity, transient-tuple width, drain-contract doc gaps) with pinning tests for each.
- Feature remains fully dormant: `settle_outbox_enabled=False`, no route calls the primitive.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #119, opened 2026-07-04
Durable settle outbox — Increment 4: enqueue-at-settle + drain endpoint

jperla:
Final increment of docs/design/durable-settle-outbox.md (§5.4, §6, §7, §8, §11). Increments 1-3 (#113, #116, #118) are merged. This wires the outbox live — still **default-off** (`settle_outbox_enabled=False`); the flag-off settle path is byte-identical (tested).

**What's in it**
- **Enqueue-at-settle** (§5.4): flag-gated durable INSERT of the frozen intent (origin/cost/validated body) before the inline finalize, 60s grace to remove the inline-vs-drain race, done-mark **only on truthy finalize** — on False the row stays pending on purpose so the drain can distinguish charged replay from reaper-free lost charge (§3). Enqueue/mark failures never fail the settle.
- **Drain endpoint** `POST /internal/gateway/settle-outbox/drain` (internal token, threadpool, limit≤500): claims due rows, applies via `apply_frozen_settle`, resolves per the §6 contract (done / force-dead+ALERT / park-no-attempt / backoff), per-row isolation, `claimed/outcomes/recovered_micro/purged` response.
- **Storage**: `park()` (attempts unchanged), `mark(force_dead=True)`, in-SQL lease fence on mark, and `purge_done()` retention — done rows >30d deleted by the drain; pending/dead/release_approved never auto-deleted (they freeze holds).
- **Fake**: strict `_require_pred` modeling for all new SQL + the MF6 follow-up anchors on the typed claim/release DML.
- **Tests**: 17 new across unit/functional/integration, including the end-to-end §1 scenario: enqueue → inline crash → reaper frozen by the Increment-2 guard → drain recovers the exact frozen cost.

**Verification**
- Gates green ×3 independently: ruff, mypy (162 files), pytest 1272 passed / 33 skipped.
- Two adversarial multi-agent verification passes (3 lenses each) drove two internal fix rounds before this PR: retention story, enqueue grace, drain robustness (clamp/threadpool/per-row isolation), warning precision (sibling self-match, zero-cost gate), vacuous fake predicate re-anchor, and lease-fence + resolver-branch coverage.

**After merge (operator steps, §8)**: apply the DDL, flip `TR_SETTLE_OUTBOX_ENABLED=true` (authorized), schedule the drain, monitor.

### DIFF
--- a/docs/design/durable-settle-outbox.md
+++ b/docs/design/durable-settle-outbox.md
@@ -164,6 +164,10 @@ the enqueue INSERT commits** — a crash between receiving the POST and the enqu
 still relies on the enclave re-delivering; quantify recovery as "finalize-after-
 enqueue failures", not "all losses".
 
+Flag-on hot-path cost: two additional sequential Spanner commits on settle
+(enqueue before finalize, done-mark after finalize), accepted for the gated
+cohort.
+
 ## 6. Drain — apply the FROZEN amount via a narrow primitive (fixes MF4/MF5/SF7)
 
 `POST /internal/gateway/settle-outbox/drain?limit=N` (internal-token auth),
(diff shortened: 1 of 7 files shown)


jperla:
codex review: PASS (1 round)

"The live wiring preserves the flag-off settle path, leaves rows pending on ambiguous finalize=False, resolves drain outcomes according to the documented contract, fences marks by lease, isolates live-billing failures, and purges only old done rows."

Pre-review hardening: two adversarial multi-agent verification passes (6 lenses total) drove two internal fix rounds before this PR — retention purge, 60s enqueue grace, drain clamp/threadpool/per-row isolation, warning precision, fake predicate re-anchoring, lease-fence + resolver-branch coverage. Gates verified independently 3× (1272 passed).

PR #120, opened 2026-07-04
Flip settle outbox on in prod (config-as-code) + operator runbook

jperla:
**Merging this PR is the production flip** — `TR_SETTLE_OUTBOX_ENABLED=true` lands in rollout.sh's ENV_VARS, and the deploy pipeline (CI gate → --no-traffic revisions → staged_traffic ramps → watchdog canaries with automatic traffic rollback) performs the rollout.

Context: all four outbox increments are merged (#113 #116 #118 #119); DDL applied 2026-07-04; reaper guard armed; drain scheduler exists paused (resumed post-verification). The earlier bare `gcloud run services update` flip attempt was a silent no-op because traffic is pinned to named revisions — the new runbook section (docs/runbook.md#settle-outbox) documents the pin gotcha, post-flip verification, the drain outcome cheat-sheet, and normal + emergency rollback.

Flag-off settle path remains byte-identical if reverted. Gates: bash -n, ruff, mypy, pytest 1272 passed.

Authorized by Joseph.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #122, opened 2026-07-05
Wire the reservation reaper into the settle-outbox drain

jperla:
The last piece of the settle-outbox system: `reap_expired_reservations` has never had a production caller — 2,636 expired holds currently leak reserved credit (one hot workspace alone holds 521). The drain (already scheduled */5) now reaps up to 200 unguarded expired holds per tick after row resolution, reports `reaped` in its response, and logs the count.

Safety: the Increment-2 guard is the interlock — pending/dead intent rows freeze their holds (in-txn re-check), and with the outbox live (#120) every fresh settle intent is enqueued before the inline finalize, so the §3 lost-charge race is structurally closed. Backlog drains in ~1 hour of ticks; steady state is far below the limit.

Tests include the full §3/§4 interplay through the drain endpoint: guarded hold frozen with zero attempts burned during a typed-store outage → recovery books the exact frozen cost → subsequent drains touch nothing. Gates: ruff, mypy, pytest 1282 passed.

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -491,9 +491,10 @@ gcloud scheduler jobs resume trusted-router-settle-outbox-drain \
 
 Every 5 min it POSTs
 `/v1/internal/gateway/settle-outbox/drain?limit=100` with the internal-token
-header and returns `{claimed, outcomes, recovered_micro, purged}`. It also
+header and returns `{claimed, outcomes, recovered_micro, purged, reaped}`. It also
 purges `done` rows older than 30 days; it never purges `pending`, `dead`, or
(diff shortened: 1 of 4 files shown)


jperla:
codex review: PASS (1 round)

"The drain-level reap uses the existing guarded reaper path, and the changed tests/fake preserve the pending/dead outbox guard behavior without weakening the expiry predicate modeling."

Gates verified independently: ruff, mypy, pytest 1282 passed. This gives reap_expired_reservations its first production caller — the ~2.6k-hold reserved-credit backlog drains at 200/tick behind the Increment-2 guard.

PR #123, opened 2026-07-05
Record the 2026-07-04/05 rollout learnings in runbook and comments

jperla:
Docs/comments only, zero behavior change. Captures tonight's operational learnings where the next operator will find them:

- New runbook section `#authorize-deadlock-burst`: triage playbook for Sentry Aborted/wounded bursts on authorize — check deploy/DDL churn first, then hot-tenant profiling via the conflict-key prefix; bursts self-resolve; shard spreading is the chronic fix. Dated receipts, no tenant identifiers.
- Settle-outbox monitoring signals: app INFO logs ship to Axiom (not Cloud Logging) — health = the open/expired-holds state query + drain tick latency, not log lines.
- `migrate_typed_counters.sh` header: apply DDL only when no deploy is rolling (2026-07-04 receipt).
- Runbook index completed (three sections were unlisted).

Gates: ruff, mypy, pytest 1282 passed, bash -n.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #124, opened 2026-07-05
Ship trusted_router INFO logs to Axiom via a scoped package-logger level

jperla:
Found during PR #123's review chain: app INFO records are dropped process-wide (root logger stays at uvicorn's WARNING; the Axiom handler never receives them). One scoped fix in `init_axiom`: set the `trusted_router` package logger to the resolved `axiom_log_level` (default INFO) — our operational lines (reaped counts, recovered charges) now ship to Axiom, third-party INFO stays gated at root, root level untouched, zero change when Axiom is disabled.

Tests assert the package-logger level, that trusted_router INFO reaches the handler while third-party INFO does not, with clean logging-state teardown. Runbook + drain comments updated to the corrected routing truth. Gates: ruff, mypy, pytest 1284 passed.

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -476,11 +476,11 @@ state; pending rows are in-flight or crash-orphaned and freeze their holds by
 design. Replayed settles (`already_settled`) never enqueue, so an empty table
 under replay-only traffic is normal.
 
-Verify there are no alert lines — in AXIOM, not Cloud Logging (app
-WARNING/ERROR ships to Axiom only; see Monitoring signals below): search the
-`trusted-router-logs` dataset for `"ALERT settle outbox"`. Equivalent
-state-based check that needs no log access at all — dead rows are the
-alert-worthy terminal state:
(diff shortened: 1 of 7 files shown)


jperla:
codex review: PASS (3 rounds)

- Round 1: FAIL — P1: the INFO unlock would flood Sentry's logs floodgate (LoggingIntegration left sentry_logs_level at INFO); P2: axiom-py ships record.args raw, which the key-based scrub can't see (email bodies with verify tokens, inquiry PII). Fixed: sentry_logs_level=WARNING; scrub filter collapses formatting and regex-redacts kv-secrets + emails.
- Round 2: FAIL — P2: TR_AXIOM_LOG_LEVEL above WARNING would suppress app warnings from Sentry (fixed: min(resolved, WARNING) cap); P2: the format-failure path still leaked raw args (fixed: args cleared unconditionally).
- Round 3: PASS — "The package logger cap and malformed-format args clearing are present and covered by tests."

Four real leak/starvation paths found and closed before a single INFO record ships. Security-contract tests pin the PII behavior.

PR #125, opened 2026-07-06
Emit a per-settle timing-breakdown INFO line (latency investigation)

jperla:
Production settles record 1-9s request latencies (predating the outbox work; the inline broadcast drain is OFF in prod, so the original suspect is exonerated). This adds one INFO line per full settle/refund with auth/enqueue/finalize/mark/total milliseconds — request-log latency minus total_ms isolates Cloud Run queue + transport time, deciding between an env fix (concurrency) and a code fix (a slow section). Replays emit nothing. Ships via #124's scoped INFO unlock; the scrub filter collapses args before anything leaves the process. Zero behavior change; two caplog tests. Gates: ruff, mypy, pytest 1290 passed.

### DIFF
--- a/src/trusted_router/routes/internal/gateway.py
+++ b/src/trusted_router/routes/internal/gateway.py
@@ -14,6 +14,7 @@ import hashlib
 import json
 import logging
 import uuid
+from time import perf_counter
 from typing import Any
 
 from fastapi import APIRouter, BackgroundTasks, Request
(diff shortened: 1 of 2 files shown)


jperla:
codex review: PASS (1 round)

"The timing instrumentation preserves the settle/refund/replay/error behavior, measures the named sections according to the spec, uses scrub-safe scalar args, and the added tests pin successful emission plus replay suppression."

Next: read an hour of prod 'settle timing' lines from Axiom against request-log latencies — request latency minus total_ms isolates Cloud Run queue time and decides the 1-9s fix.

PR #127, opened 2026-07-06
Add retry-attempt count to typed finalize timing (contention attribution)

jperla:
Prod `typed finalize timing` shows `spanner_ms` p50 977ms / p90 5.76s — but that number bundles wound-retry contention and baseline multi-region commit latency. This surfaces the retry helper's winning attempt count onto the timing line (`attempts=N`) so we can bucket spanner_ms by attempts and separate the two.

Caveat documented: counts only the outer `run_in_transaction_with_retry` retries — Spanner's own internal Aborted retries are invisible — so `attempts>1` is a definitive severe-contention signal, while `attempts==1` doesn't rule out contention Spanner absorbed internally. Optional `attempts_out`, default None = zero behavior change. Tests cover the retry counter and the extended timing line. Gates: ruff, mypy, pytest 1292 passed.

### DIFF
--- a/src/trusted_router/storage_gcp.py
+++ b/src/trusted_router/storage_gcp.py
@@ -1072,11 +1072,15 @@ class SpannerBigtableStore:
                 index_ms = (time.perf_counter() - index_start) * 1000
             # Splits the settle-path finalize_ms hotspot (2026-07-05 investigation)
             # into Spanner-txn vs Bigtable-index time.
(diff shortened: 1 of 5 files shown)


jperla:
codex review: PASS (2 rounds). Round 1 flagged missing retry-counter coverage; the new tests/test_storage_gcp_io.py already had the winning-attempt cases and commit 2 added the exhausted-retries case. Merging to read the attempts distribution against spanner_ms in prod.

PR #129, opened 2026-07-06
Control-plane parity: honor max_completion_tokens / max_output_tokens

jperla:
Companion to the enclave fix for the publicly-reported truncation bug. quill-router read only `max_tokens` for the output cap, so clients using the modern `max_completion_tokens` spelling got an under-sized billing reservation and an under-capped provider request on the local/test + meta-routing paths. Adds one shared `resolve_max_output_tokens` accessor (precedence max_tokens > max_completion_tokens > max_output_tokens) used by the adapter, provider payloads, and the authorize estimate. Focused alias tests; gates ruff/mypy/pytest 1302 passed. Behavior identical for existing max_tokens clients.

### DIFF
--- a/src/trusted_router/adapter.py
+++ b/src/trusted_router/adapter.py
@@ -18,16 +18,29 @@ from typing import Any
 
 from trusted_router.errors import api_error
 
+_OUTPUT_TOKEN_FIELDS = ("max_tokens", "max_completion_tokens", "max_output_tokens")
+
 
 def resolve_max_output_tokens(body: Mapping[str, Any]) -> int | None:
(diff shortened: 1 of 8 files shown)


jperla:
codex review: PASS (4 rounds). Rounds surfaced real edge cases on the billing-input path: (1) _estimate_reserve still read only max_tokens (alias-only under-reserve); (2) output_estimate precedence; (3) non-integer aliases 500'd on /messages+/responses; (4) streaming validated the alias mid-stream. All fixed with tests; final PASS confirms no mid-stream coercion paths remain. Merging on CI green.

PR #156, opened 2026-07-12
billing: dark credit-row sharding plumbing (increment 1)

jperla:
Implements Increment 1 from #155.

Scope:
- additive tr_reservation.credit_shard migration with rolling fallback to legacy ws_shard
- CreditAccount shard_count default 1
- reserve records the selected credit shard; settle/refund release that recorded shard
- direct grants split deltas evenly across configured shards, remainder on shard 0
- typed snapshots, drift comparison, and invariant audit aggregate credit shards
- generic absolute-value credit mirror refuses to overwrite sharded sub-budgets
- no shard selection, scan, rebalance, operator activation, or production workspace sharding

Dark behavior:
- every existing and new workspace remains shard_count=1
- authorize still selects shard 0
- external gateway/API response shapes are unchanged
- existing reservations with no credit_shard settle through ws_shard fallback

Rollout requirement:
1. Apply scripts/deploy/migrate_typed_counters.sh before deploying application code.
2. Deploy the control plane region by region.
3. Keep shard_count=1 everywhere.
4. Run the typed invariant auditor after each region.

Verification:
- 1423 passed, 33 skipped
- uv run ruff check .
- uv run mypy
- bash -n scripts/deploy/migrate_typed_counters.sh
- git diff --check

This intentionally does not mix in the retry-budget or event-loop emergency patches.

### DIFF
--- a/scripts/deploy/migrate_typed_counters.sh
+++ b/scripts/deploy/migrate_typed_counters.sh
@@ -120,6 +120,7 @@ if table_exists tr_reservation; then log "tr_reservation exists, skip"; else
     workspace_id STRING(64),
     key_hash STRING(64),
     ws_shard INT64,
+    credit_shard INT64 NOT NULL DEFAULT (0),
     key_shard INT64,
     credit_reserved_micro INT64,
     key_reserved_micro INT64,
@@ -180,6 +181,11 @@ ensure_column tr_key_limit week_usage  "INT64 DEFAULT (0)"
 ensure_column tr_key_limit week_start  "TIMESTAMP"
 ensure_column tr_key_limit month_usage "INT64 DEFAULT (0)"
 ensure_column tr_key_limit month_start "TIMESTAMP"
+
+# Credit-row sharding increment 1. Existing reservations retain ws_shard=0;
+# old rows read NULL for the additive column and the application falls back to
+# ws_shard. New rows write both during the rolling-compatibility window.
+ensure_column tr_reservation credit_shard "INT64 DEFAULT (0)"
 
 # ── Durable settle outbox (docs/design/durable-settle-outbox.md) ────────────
 # Recovers completed-but-settle-lost charges. Additive + guarded; safe to apply
--- a/src/trusted_router/storage_gcp.py
+++ b/src/trusted_router/storage_gcp.py
@@ -50,7 +50,11 @@ from trusted_router.storage_gcp_codec import (
     normalize_email as _normalize_email,
 )
 from trusted_router.storage_gcp_counter_dml import insert_entity_dml_at, update_entity_body_dml
-from trusted_router.storage_gcp_counters import UNSHARDED
+from trusted_router.storage_gcp_counters import (
+    UNSHARDED,
+    credit_shard_count,
+    distribute_credit_amount,
+)
 from trusted_router.storage_gcp_counters import mirror_delete as mirror_counter_delete
 from trusted_router.storage_gcp_counters import mirror_write as mirror_counter_write
 from trusted_router.storage_gcp_custom_models import SpannerCustomModels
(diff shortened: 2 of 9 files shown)


src/trusted_router/storage_gcp_counters.py
```diff
@@ -110,6 +143,8 @@ def mirror_write(writer: Any, kind: str, entity_id: str, value: Any, commit_ts:
     writer, so the mirror commits atomically with it. No-op for other kinds.
     """
     if kind == "credit":
+        if credit_shard_count(value) != 1:
```
jperla, line 146: ✅ **This is the load-bearing safety line and it is correct.** Skipping the absolute-value mirror once `shard_count != 1` is exactly what prevents the 2026-06-25 clobber class (replaying the whole JSON total onto shard 0 would multiply a sharded budget). Verified the corollary holds: `credit_workspace_typed_direct` distributes deltas per-shard and writes the JSON warm-copy via direct DML (not `_write_entity_tx`), so it never re-enters this generic mirror — no double-apply. `credit_shard_count` failing closed on bad values is the right call too.

src/trusted_router/storage_gcp.py
```diff
@@ -1299,23 +1311,39 @@ def typed_key_usage(self, key_hash: str) -> dict[str, Any] | None:
...
+        when the typed tables are off or a one-shard row is not seeded. A
+        configured multi-shard workspace fails closed if any active row is
+        missing; falling back to stale JSON usage would overstate availability.
+        """
         if not getattr(self, "_counter_mirror_enabled", False):
             return None
         pt = self._param_types
-        with self._database.snapshot() as snapshot:
+        with self._database.snapshot(multi_use=True) as snapshot:
+            account = self._read_entity_from(snapshot, "credit", workspace_id, CreditAccount)
```
jperla, line 1325: **[P3] Read amplification that is live immediately (not dark).** `typed_credit_snapshot` now does an extra `credit` entity read before the shard scan — for every call, including the shard_count=1 case that is 100% of traffic today. So unlike the rest of this PR, this path changes behavior for all workspaces now: one point-read of shard 0 became an entity read + a ranged shard scan under a `multi_use` snapshot. It is a display/balance path (not hot authorize/settle), so not urgent, but consider caching `shard_count` on the workspace or reading it from the already-loaded principal to avoid the extra round-trip. Worth a quick check that no hot path calls this per-request.

scripts/deploy/migrate_typed_counters.sh
```diff
@@ -181,6 +182,11 @@ ensure_column tr_key_limit week_start  "TIMESTAMP"
 ensure_column tr_key_limit month_usage "INT64 DEFAULT (0)"
 ensure_column tr_key_limit month_start "TIMESTAMP"
 
+# Credit-row sharding increment 1. Existing reservations retain ws_shard=0;
+# old rows read NULL for the additive column and the application falls back to
+# ws_shard. New rows write both during the rolling-compatibility window.
+ensure_column tr_reservation credit_shard "INT64 DEFAULT (0)"
```
jperla, line 188: **[P3] Nullability diverges by install history.** Fresh installs create `credit_shard INT64 NOT NULL DEFAULT (0)` (the CREATE block above); existing installs get `INT64 DEFAULT (0)` (nullable) via this `ensure_column`. Functionally fine — the app tolerates NULL and falls back to `ws_shard` — and nullable-on-existing is the correct choice (a NOT NULL add would force a synchronous backfill on a large table). Just call it out in a comment so a future reader does not "fix" the divergence into a NOT NULL add and trigger that backfill. Optionally backfill + tighten to NOT NULL in a later increment once the fallback window closes.

jperla:
## Review summary (Claude Fable) — APPROVE. Clean, safe, well-tested dark increment.

This faithfully implements increment 1 of the sharding handoff and I could not find a correctness defect. The two properties that make or break a dark increment both hold:

**1. Byte-identical at shard_count=1 (100% of traffic).** Authorize records `credit_shard` defaulting to `UNSHARDED`; settle/refund release the recorded shard with a `ws_shard` fallback for pre-migration rows; grants distribute via `distribute_credit_amount` which returns `(amount,)` for one shard; `typed_credit_snapshot` sums a single row. Verified the pre-migration reservation path and idempotent-replay-keeps-shard.

**2. Cannot corrupt a sharded sub-budget.** The absolute-value mirror is skipped once `shard_count != 1` (inline-confirmed), `credit_shard_count` fails closed on bad values, and `credit_workspace_typed_direct` writes per-shard deltas + the JSON warm-copy via direct DML (never re-entering the mirror). The invariant auditor and `compare()` now aggregate per-shard credit holds — so the tripwire that will guard the eventual jay flip is already shard-aware.

Migration is additive (`ensure_column`, nullable-default) with the rolling `ws_shard` fallback documented; the 13-test matrix covers additive schema, legacy default, fail-closed shard_count, grant remainder-on-0, authorize/settle shard round-trip, snapshot sum + fail-closed, mirror non-clobber, and shard-aware compare/audit. CI green (1423 passed).

Left three inline notes: a ✅ confirmation on the mirror-skip safety line, and two **P3s** — `typed_credit_snapshot` adds a read for all workspaces immediately (not dark), and the additive-column nullability diverges by install history (fine, just document it).

**Merge/deploy note:** this is independent of and correctly excludes the emergency retry-budget/event-loop patches. It is dark, so it does not itself relieve the incident — but it is also safe to land whenever, provided the migration (`migrate_typed_counters.sh`) runs before the code per the PR's rollout steps. Nothing here blocks or is blocked by the acute fix.

scripts/deploy/migrate_typed_counters.sh
jperla: Addressed in the stacked follow-up: the migration now explicitly documents that existing databases intentionally remain nullable to avoid a synchronous backfill, while fresh installs are NOT NULL, and that tightening belongs in a later post-backfill migration after the rolling fallback is retired.

src/trusted_router/storage_gcp.py
jperla: Investigated the call graph: this snapshot feeds auto-refill decisions as well as display. The Increment 2 shard-count cache is deliberately allow-stale for authorize safety, but using it here could understate balance immediately after a split and trigger an early card charge. I am therefore keeping the exact same-snapshot account read here and documented that constraint in the stack. #158 caches the genuinely hot authorize lookup; a future optimization here needs an exact typed config projection or an already-loaded exact count, not the allow-stale cache.

jperla:
Superseded by the complete reviewed rollup merged in #162. The focused commit remains preserved in #162's history and review record.

PR #157, opened 2026-07-12
Request tagging and OpenRouter attribution compatibility

jperla:
## Summary

Adds AWS-style request tags and OpenRouter-compatible attribution across authorization, idempotency, activity, generation metadata, API-key defaults, and public documentation.

## Review fixes

- Effective tags are frozen at authorization and settlement remains money-safe.
- Activity grouping reports truncation and caps cardinality.
- Tag filters report their bounded recent-window scope.
- Portable whitespace validation excludes Unicode line and paragraph separators.
- Authorize responses advertise request_metadata_version: 1 for safe gateway rollout.
- User and session compatibility limits match the existing 256-character contract.
- Explicit empty tags preserve the legacy idempotency fingerprint across deployment.

## Deployment order

Merge and deploy this control-plane PR first. Verify production authorize responses include request_metadata_version: 1 before merging Lore-Hex/quill-cloud-proxy#21. Both PRs remain draft while the authorize and settle incident is monitored.

## Verification

- 1,514 tests passed, 33 skipped
- Ruff passed
- mypy passed
- GitHub coverage and browser smoke passed

Design: docs/design/request-tagging.md


src/trusted_router/storage_activity.py
```diff
@@ -37,21 +37,28 @@ def filter_generations(
...
-    grouped: dict[tuple[str, str, str], dict[str, Any]] = {}
+def summarize_activity(
+    generations: Iterable[Generation], *, group_by_tag: str | None = None
+) -> list[dict[str, Any]]:
+    grouped: dict[tuple[str, str, str, str | None], dict[str, Any]] = {}
     for gen in generations:
         day = gen.created_at[:10]
-        key = (day, gen.model, gen.provider_name)
+        grouped_tag_value = gen.tags.get(group_by_tag) if group_by_tag else None
+        key = (day, gen.model, gen.provider_name, grouped_tag_value)
```
jperla, line 96: **[P2] Tag group-by repeats the #142 truncation bug.** `summarize_activity` sums over a bounded scan (`activity()` fetches `limit=5000`) but the result carries no `truncated` signal, and tag grouping has no cardinality cap. A workspace with >5000 recent generations gets tag-grouped **cost sums that are silent lower bounds** — a partial sum presented as exact, the exact class of bug we just fixed for usage graphs in #142. Also `group_by=tag:request-id` produces unbounded groups. Please (a) surface `truncated: true` when the scan hits the cap, and (b) cap group cardinality (top-N + an `other` bucket + `groups_truncated`). This is cost-allocation data customers will trust; wrong-low is a real problem.

src/trusted_router/storage_gcp_generations.py
```diff
@@ -153,11 +157,22 @@ def activity(
         *,
         api_key_hash: str | None = None,
         date: str | None = None,
+        tag_key: str | None = None,
+        tag_value: str | None = None,
+        group_by_tag: str | None = None,
     ) -> list[dict[str, Any]]:
         rows = self._activity_generations(
             workspace_id, api_key_hash=api_key_hash, date=date, limit=5000
```
jperla: **[P3] Tag filtering runs AFTER the recent-window scan, not inside the Bigtable range.** `activity()` pulls the 5000 most-recent rows, then `filter_generations` narrows by tag in memory. For a rare tag on a busy workspace, those 5000 rows can contain zero matches even though matches exist just outside the window → the user sees "no data" for a tag that is real. For now, document tag filters as best-effort over the recent window and point at the future rollup (design §5). The durable answer is the activated-tag rollup, not a bigger scan.

src/trusted_router/request_tags.py
```diff
@@ -0,0 +1,92 @@
...
+    """Compare a supplied settle map with authorization-frozen tags."""
+    return validate_tags(given) == dict(frozen)
+
+
+def _portable_tag_text(value: str) -> bool:
+    for char in value:
+        if char in _PORTABLE_PUNCTUATION:
+            continue
+        category = unicodedata.category(char)
+        if category[0] in {"L", "N", "Z"}:
```
jperla: **[P3] `Z` allows more than "spaces".** Unicode category `Z` includes line/paragraph separators (`Zl`/`Zp`, e.g. U+2028/U+2029), not just spaces. AWS's portable set is letters, numbers, and spaces (`Zs`). Tighten to `Zs` (or an explicit space allowlist) so a tag value cannot smuggle a line separator. Not security-critical — tags never reach prompts/providers — but it keeps the character contract honest and matches the AWS reference the doc cites.

jperla:
## Review summary (Claude Fable) — money-safe, hold merge until the incident clears

The tag logic is **correct and money-safe**; it incorporated every blocking item from the design review:
- Settle never rejects on tag mismatch — logs counts-only, builds the generation from `authorization.tags` (frozen at authorize); settlement stays a pure function of money state. ✅
- Idempotency fingerprints the **request** tags (not the merged set), drops the key when tagless (byte-identical to today's fingerprint → in-flight retries survive the deploy), and `sort_keys` canonicalizes so reordered tags don't 409. ✅
- 4 KiB aggregate cap, 50-tag cap, merged-over-50 error with counts, frozen tags persist on the production generation. ✅

Three code items left as inline comments: **[P2]** activity tag group-by has no truncation flag / cardinality cap (the #142 partial-sum-shown-as-exact class), **[P3]** tag filter runs after the recent-window scan, **[P3]** `_portable_tag_text` allows Unicode `Z` (line separators) not just `Zs`.

**[P3 / follow-up]** The never-forward-to-provider guarantee (`tags`/`trace`/`session_id`/`user`) is **not proven by this control-plane PR** — the provider invoke is enclave-side. The §11 trust-boundary test belongs in the enclave follow-up PR and must exist before the feature is called done.

**Operational — the one that matters most:** please do **not** merge this during the active authorize/settle latency incident. It reworks the exact hot path (every authorize now runs tag validation + a changed fingerprint build). Land the acute retry-budget fix first, confirm the affected workspace is stable, then merge tagging. Nothing here is wrong; the timing is the risk.

jperla:
Fable review follow-up: all three findings are fixed in the updated head. Activity responses now expose scan and group truncation, tag group-by uses top 100 plus __other__, tag filters explicitly report recent-window scope, and whitespace validation allows Unicode space separators only. The attested provider-boundary coverage is in Lore-Hex/quill-cloud-proxy#21. Full control CI is green, including 1,439 tests, coverage, and browser smoke. This PR remains draft pending incident stability.

src/trusted_router/storage_gcp_generations.py
jperla: Documented and surfaced in 6f698f7. Filtered activity responses now report tag_filter_scope=recent_window plus truncation and scan-limit metadata, and the public tagging documentation states that filtering operates over a bounded recent or date-scoped scan pending durable activated-tag rollups.

src/trusted_router/request_tags.py
jperla: Fixed in 6f698f7. Portable tag whitespace now accepts Unicode category Zs only. Regression coverage explicitly rejects U+2028 while preserving ordinary Unicode space separators.

src/trusted_router/storage_activity.py
jperla: Fixed in 6f698f7. ActivityResult now exposes truncated, scanned, scan_limit, and groups_truncated. Tag grouping retains the 100 most frequent values and folds all remaining values into __other__. Regression coverage includes a 5,001-row truncated scan and 101 tag values.

src/trusted_router/schemas.py
```diff
@@ -192,6 +194,10 @@ class GatewayAuthorizeRequest(_Lenient):
     session_id: str | None = Field(default=None, max_length=256)
     trace: dict[str, Any] | None = None
     metadata: dict[str, Any] | None = None
+    tags: dict[str, Any] | None = None
+    app: str | None = Field(default=None, max_length=120)
+    http_referer: str | None = Field(default=None, max_length=2048)
+    app_categories: list[str] | None = None
```
jperla, line 200: **[P2] Attribution/tag limits here diverge from both the design doc and the enclave (#21), so the two services disagree on what's valid.**
- `user`/`session_id` are `max_length=256` (lines 193-194), but design §2.1 and the enclave both cap them at **128** (`MaxAttributionUserRunes=128`). A 200-char `user` is accepted here, rejected at the enclave.
- `app_categories` (this line) has **no** validation — the enclave caps at **2** (`MaxAttributionCategories=2`).
- `trace` (line 195) has no size/depth bound; design §2.1 and the enclave enforce ~8 KiB / bounded depth.

Whichever service the value reaches first decides the outcome, so identical input yields inconsistent 400s. Align these limits with the enclave contract (128 / ≤2 categories / bounded trace).

src/trusted_router/routes/internal/gateway.py
```diff
@@ -162,10 +188,16 @@ async def authorize_gateway(
     request_idempotency_key = _gateway_idempotency_key(request, body) or str(
         uuid.uuid4()
     )
+    fingerprint_body = dict(body_dict)
```
jperla, line 210: **[P3] Empty-tags fingerprint change can 409 an idempotent retry that spans the router deploy.** Here the fingerprint drops `tags` when empty (`fingerprint_body.pop("tags", None)`), but the *current* lenient router (`extra="allow"`) folds a client-sent `"tags": {}` into `body_dict` and thus into the fingerprint. So a client sending an explicit empty `tags: {}` and retrying with the same idempotency key across the #157 deploy gets a 409 (fingerprint mismatch) instead of a replay. Narrow, but worth normalizing empty-vs-absent tags identically on both sides of the flip (or noting it in the rollout).

jperla (commented):
**Follow-up review (Claude Fable), cross-checked against the enclave companion Lore-Hex/quill-cloud-proxy#21.** The tag logic itself remains correct and money-safe (the three earlier findings are fixed in this head). These are new findings on the control-plane ↔ enclave *contract*.

**Deploy ordering:** this PR must merge and be serving **before** the enclave #21, per the design doc rollout. Otherwise the enclave enforces the tag/attribution contract (400s on invalid input) while the current lenient router silently accepts-and-drops tags (200 with no attribution recorded) — indistinguishable from success. Also: this PR is currently CONFLICTING with main and needs a rebase.

jperla:
Fable follow-up is complete on head 7a5acac. The branch is rebased onto current main, preserving the new credit and key sharding fields alongside request tags. Production rollout capability version 1 is included. Local verification: 1,513 passed and 33 skipped, plus Ruff and mypy. GitHub CI is green, including coverage and browser smoke. This remains draft and must deploy before the enclave companion.

src/trusted_router/routes/internal/gateway.py
jperla: Fixed in 69c8cf2. The fingerprint now distinguishes an absent tags field from an explicitly supplied empty object, matching the pre-tagging lenient router. Ordinary tagless requests remain byte-compatible, while a tags: {} retry spanning deployment replays instead of returning 409. The regression computes the exact legacy fingerprint, asserts the stored authorization matches it, then verifies replay.

src/trusted_router/schemas.py
jperla: Resolved across control 69c8cf2 and gateway 3bd2262. We chose the existing production-compatible 256-character user/session contract and aligned the enclave to it. The control route calls validate_request_attribution before routing or reservation; that same validator enforces at most two categories, 8 KiB trace JSON, depth 8, and 256 aggregate keys/items. The schema remains shape-lenient intentionally so these failures use the stable OpenAI-shaped 400 contract instead of a generic Pydantic 422. Regression tests cover all limits.

PR #174, opened 2026-07-14
synthetic: give modern reasoning models probe headroom

jperla:
## Why
Debugging the `/leaderboard` error columns: ~23% of probe failures were `probe_config_error` (already excluded from uptime, but shown in "Config excluded" and cluttering per-model error rates). Root cause: the rotation probe's `_rotation_max_tokens` allowlist is stale — the entire modern reasoning lineup falls through to the **16-token default**. On a "reply exactly PONG" probe these models spend the budget on hidden reasoning and the stream ends `finish_reason=length` before any content/reasoning delta, which the classifier records as a config error even though the provider is healthy.

Confirmed against live leaderboard data (2026-07-13) — all 15 `(provider, model)` pairs hitting `finish_reason=length` were reasoning models sent the 16-token default:

`gemma-4-26b`, `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite-preview`, `gemini-2.5-pro`, `qwen3.6-35b/27b`, `qwen3.5-397b/35b`, `deepseek-v4-pro`, `nemotron-3-{nano,super,omni}`, `minimax-m2.5`.

## What
- Add these families to the **existing, proven 512-token reasoning tier** (gpt-oss/glm-5 already in that tier don't fail at 512)
- **Cost stays flat**: the cheap default is untouched for non-reasoning models. Matching is deliberately scoped to exclude `gemini-2.5-flash-lite` — a non-reasoning model and our single highest-volume route — so its probe measurement isn't perturbed
- The leaderboard's per-model `probe_config_error` rate is documented in-code as the drift detector for the next reasoning family (so this doesn't silently go stale again)

## Verification
- New tests assert all 15 observed-failing pairs now get 512, and that non-reasoning models (esp. `gemini-2.5-flash-lite`) stay at 16
- Full synthetic-monitoring suite: **65 passed**; ruff clean

Left out of scope per discussion: the enclave-502 regional issue (was billing, already fixed) and the leaderboard gateway-502 classifier change (revisit with more data).

### DIFF
--- a/src/trusted_router/synthetic/probes.py
+++ b/src/trusted_router/synthetic/probes.py
@@ -956,25 +956,64 @@ async def provider_rotation_probe(
     )
 
 
+# Model-id substrings for models that reason/think before emitting visible
+# content. They must get REASONING_MAX_TOKENS headroom in the rotation probe:
+# with the cheap default they exhaust the budget mid-thought and stream ends
+# with finish_reason=length before any content/reasoning delta, which the
+# classifier records as `probe_config_error` — inflating the model's error
+# column on the public /leaderboard even though the provider is healthy.
+#
+# DRIFT DETECTOR: the leaderboard's per-model `probe_config_error` rate is the
+# signal that a new reasoning family has shipped and needs adding here. As of
+# 2026-07-13 the whole modern reasoning lineup (Gemini 2.5+/3.x, Qwen3.5/3.6,
+# Gemma-4 thinking, Nemotron-3, DeepSeek V4 Pro, MiniMax M) was silently
+# defaulting to 16 and failing. Non-reasoning models keep the cheap cap — the
+# extra tokens are only spent where a model actually needs to finish thinking.
+REASONING_MAX_TOKENS = 512
+_REASONING_MODEL_MARKERS = (
+    # explicit self-describing ids
+    "reasoning",
+    "thinking",
+    # OpenAI o-series + GPT-5 (also matched provider-scoped below)
+    "gpt-5",
+    # open-weight reasoners
+    "gpt-oss",
+    "glm-4.6",
+    "glm-4.7",
+    "glm-5",
+    # Gemini Pro + 3.x preview models reason server-side and emit content only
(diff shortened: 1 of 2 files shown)


jperla:
Closing — the mechanism landed in a different and better form. Recording the one part that did not, so it is not lost.

**Superseded.** Main has `_rotation_max_tokens(provider, model)` in `synthetic/probes.py`, which does what this PR's flat `REASONING_MAX_TOKENS = 512` + `_REASONING_MODEL_MARKERS` list does, but per family:

- OpenAI o1/o3/o4/gpt-5 → 512
- Gemini 2.5/3.x → **2048**, with a note that live verification on 2026-07-19 showed 512 was not enough
- gpt-oss, glm-4.6/4.7/5, nemotron, claude-fable-5, claude-sonnet-5, and any id containing `reasoning`/`thinking` → 512
- kimi-k2, grok, claude-opus → 128, everything else → 16

The Gemini case is why main's shape is better: this PR's single 512 constant would still have starved it.

**The part that did not land, unverified.** Main's function does not name four families this PR called out: Qwen3.5/3.6, Gemma-4 thinking, DeepSeek V4 Pro, MiniMax M. I evaluated the live function against the catalog and **11 entries currently resolve to the 16-token default**:

    qwen/qwen3.5-9b, qwen/qwen3.6-27b, google/gemma-4-31b-it,
    deepseek/deepseek-v4-flash, deepseek/deepseek-v4-flash-0731,
    deepseek/deepseek-v4-pro-0423, deepseek/deepseek-v4-pro-0813,
    minimax/minimax-m3, minimax, minimaxai, MiniMaxAI

I did **not** add them, because I could not verify they need the headroom: I found no `probe_config_error`/`empty_stream` signal for them in `status.json` and there is no reasoning flag on them in catalog metadata. Adding token budgets on a July assertion would be guessing.

The check is the one this PR itself names as the drift detector: the per-model `probe_config_error` rate on the leaderboard. If any of those 11 shows an inflated error column while the provider is healthy, the fix is one clause in `_rotation_max_tokens`, not a revival of this branch.

PR #197, opened 2026-07-16
Publish Prometheus, OpenPatcher, Iris, and Plato K3 upgrades

jperla:
## Summary
- publish trustedrouter/prometheus-2.0 and trustedrouter/openpatcher-g2
- publish K3 upgrades trustedrouter/iris-2.0, trustedrouter/openpatcher-s2, and trustedrouter/plato-pro-2.0
- advance rolling Prometheus, Iris, and Plato Pro aliases while preserving pinned prior versions
- keep new orchestration models Credits-only and update public docs, model pages, and agent guidance
- add catalog, routing, alias, context, discoverability, and immutability contracts

## Local verification
- uv run ruff check .
- uv run mypy
- uv run pytest -q: 1569 passed, 33 skipped
- npm run typecheck
- npm run lint
- npm run lint:css

## Deployment order
Merge and deploy the matching quill-cloud-proxy orchestration PR first so the attested gateway understands every ID before this catalog is published.


jperla:
Blocking this PR from merge for now: commit 6b3b98f adds Iris 2.0, OpenPatcher S2, and Plato Pro 2.0 catalog IDs, but the deployed enclave from quill-cloud-proxy#29 implements only Prometheus 2.0 and OpenPatcher G2. Merging #197 as-is would advertise IDs that the attested runtime cannot execute. The runtime-aligned subset is in #198; preserve this PR for a follow-up that adds and deploys the matching enclave implementations first.

PR #210, opened 2026-07-17
routing: drop GMI Credits route for kimi-k3 (404s; takes down prometheus-2.0)

jperla:
## What broke

`trustedrouter/prometheus-2.0` (and every K3 combo) is hard-down: **`moonshotai/kimi-k3` → 404 "No matching target server found for model kimi-k3"**. kimi-k3 is prometheus-2.0's synthesizer, judge, and a panel member, so the preset returns `404 fusion failed` on every call.

## Why

- Pinning `provider.order=["gmi"]` reproduces the 404 → **GMI no longer serves kimi-k3**, and the default route preferred GMI.
- Pinning `provider.order=["moonshot"]` returns `OK` through the live gateway → **the first-party route is healthy**.
- Moonshot's own API lists `kimi-k3` as served (checked directly).
- Same failure class as the 2026-07-15 nemotron/GMI entry directly above in the denylist; same fix shape (Credits-only, BYOK stays visible).

## Verification

- Built registry from this branch: gmi **Credits** endpoint for kimi-k3 dropped; first-party Credits endpoint retained; gmi BYOK untouched.
- `tests/test_catalog_routing_contracts.py`: **84/84 pass**.

## Follow-up (systemic, not in this PR)

Second incident today where one dead provider route blackholed a model with a healthy sibling route (nemotron yesterday, kimi-k3 today). A provider-sourced 404 of this shape ("no matching target server") should be **failover-eligible** in the gateway's upstream attempt loop rather than terminal.

@codex please review and deploy — unblocks the prometheus-2.0 DRACO eval.


jperla:
Closing — the condition this fixed has been reversed deliberately since.

This marks `moonshotai/kimi-k3` unserved on GMI because GMI 404'd it in July. Main now asserts the opposite: `tests/test_kimi_k3_gmi_phala.py::test_gmi_kimi_k3_is_a_verified_prepaid_route` requires `moonshotai/kimi-k3@gmi/prepaid` to exist as a verified prepaid route with specific list pricing, added in #699 (`82e56c5b` "Price GMI routes from stable list rates").

Merging this would break that contract. Refreshed against current main, it fails `test` and `test-post-cutover`:

    FAILED tests/test_kimi_k3_gmi_phala.py::test_gmi_kimi_k3_is_a_verified_prepaid_route - KeyError: 'moonshotai/kimi-k3@gmi/prepaid'
    FAILED tests/test_catalog_routing_contracts.py::test_every_catalog_model_has_integer_prices_and_valid_provider

Worth noting the PR's green tick was from a retired CI config that ran a single `test` job; the current suite is 8 checks. If GMI starts 404'ing that route again the fix is to re-cut against today's catalog, not to revive this.

PR #212, opened 2026-07-17
settle: Sentry warning on provider failovers (accumulate dead-route signal)

jperla:
Pairs with [quill-cloud-proxy#37](https://github.com/Lore-Hex/quill-cloud-proxy/pull/37). Deploy order doesn't matter (`GatewaySettleRequest` is `extra="allow"`; fields are optional).

## What

When the attested gateway settles a request that succeeded only after routing around a failing provider, emit a **Sentry warning**:

- message: `provider failover: <model> routed around <providers>` — one Sentry issue per model+provider pair, so the issue's event count is the accumulation signal ("if it happens too much, look into it")
- tags: `failover.model`, `failover.failed_providers`; context: the `provider|endpoint|error_class` list + which endpoint served
- fire-and-forget: reporting can never break settlement

## Why

The router already fails over automatically — the enclave tries every authorized candidate on any pre-output provider error. What was missing is the signal: GMI silently dropped nemotron (07-15) and then kimi-k3 (07-16) while being the preferred route, and nothing surfaced until *all* candidates failed simultaneously and synth presets went down. With this, each routed-around failure lands in Sentry days earlier; the fix is then a one-line `_PROVIDER_UNSERVED_CREDITS_MODELS` demotion.

## Tested

- gateway cache-billing, aborted-handler, core API tests pass (34/34 + 5/5)
- hook unit-verified: emits `level=warning`, correct message/tags/context; `getattr` access keeps `model_construct`-built bodies working

@codex please review and deploy (with quill-cloud-proxy#37).


jperla:
Held this back from merging — it was detaching the billing-path Spanner RPC budget, and nine green checks didn't see it.

`_report_route_fallbacks` was inserted **between** `@spanner_rpc_budget(_BILLING_PATH_SPANNER_BUDGET_SECONDS)` and `_settle_gateway_authorization`. Python permits blank lines between a decorator and its `def`, and binds the decorator to the *first* function that follows — so the budget landed on the Sentry helper and the function that writes billing state was left unguarded. Minimal repro:

```python
@deco(5)


def first(): ...
def second(): ...
# first.budget -> True,  second.budget -> False
```

It compiled, imported, and passed the whole suite, because the budget only changes behaviour under RPC pressure — nothing behavioural can observe it.

Fixed in `bcba6e13`: the helper moves above the decorator, the decorator goes back onto `_settle_gateway_authorization`. Confirmed from the AST rather than by reading — `_authorize_gateway_sync` and `_settle_gateway_authorization` carry the budget, `_report_route_fallbacks` does not.

Added `tests/test_billing_path_rpc_budget.py` so this can't recur silently. It asserts set **equality**, so it fails both if the budget detaches and if it drifts onto an extra function, plus a second test that refuses a budget decorator followed by anything other than a `def`. Verified by mutation: reintroducing the blank lines turns it red, removing them turns it green.

The Sentry accumulation logic itself is unchanged and looks right — the failover signal, the grouping by model+providers, and the swallow-everything guard around the report are all fine. 291 gateway/settle tests pass, ruff and mypy clean.

PR #339, opened 2026-07-29
Fix window-floor and retention-TTL interleavings in the recent Spanner changes

jperla:
Review of the recent Spanner work (#318 key sharding, #319 expired legacy holds, #320 preserve live holds, #314 retention) found **four concrete defects**. None is silent money loss in steady state, but all are reachable and none was covered by the landed tests.

The reshard trio's money conservation, the live-hold settle test, #319's conservatism, and the retention migration's safety checks all held up under review — these are the gaps around the edges.

### 1. Reshard captured window floors outside its transaction
`storage_gcp_key_shard_admin.py` computed `window_floors(utcnow())` before `def txn`, and Spanner abort-retries re-ran the body with that same stale value. If a UTC day/week/month boundary crossed in between, the reshard kept the new window's usage but rewrote every shard's window start to the **old** floor — so the next limit check saw an expired window, lazily reset usage to zero, and a **hard** daily/weekly/monthly cap (hard since #280) could be exceeded for the rest of that window.

**Fix:** derive floors per transaction attempt.

### 2–4. Retention TTL could be armed while an outbox intent was outstanding
`tr_reservation` / `tr_gateway_authorization` carry `terminal_at` with a 30-day row-deletion policy. `pending`/`dead` outbox rows are themselves TTL-immune and are the sole lost-charge signal — but the records they *refer to* were not protected:

- **The reaper arms `terminal_at`** when it free-releases, so a settle that enqueued just after losing that race left a correctly-frozen `dead` row whose reservation/authorization disappeared 30 days later.
- **Sibling intents:** the PK is `(authorization_id, intent_kind)`, so settle and refund coexist by design — but `mark(done=True)` armed retention unconditionally, letting the first intent to finish start the clock on records the sibling still needed.
- **A winning claim** arms `terminal_at` at settle time, so a row that later went `dead` kept a 30-day fuse on the very evidence its freeze exists to preserve.

**Fix:** enqueue (and refresh) clear `terminal_at` on both referenced records; `mark(done)` arms only when no sibling remains `pending`/`dead`; any non-`done` outcome disarms. Net invariant: **an outstanding outbox intent always implies TTL-ineligible records.**

### Tests
Adds the four regression tests these interleavings lacked — boundary-crossing reshard, reaper-then-late-settle, settle+refund siblings, dead-row disarm. Each was verified to **fail without its fix**.

Gates: `ruff` clean · `mypy` 190 files · `pytest` 2113 passed / 61 skipped.

### DIFF
--- a/src/trusted_router/storage_gcp_authorize.py
+++ b/src/trusted_router/storage_gcp_authorize.py
@@ -24,7 +24,6 @@ import logging
 import threading
 import time
 import uuid
-import weakref
 from collections.abc import Callable
 from typing import Any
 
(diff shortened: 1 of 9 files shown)


jperla:
codex review: PASS (6 rounds)

Round-by-round, each finding fixed with a regression test verified to fail without it:
- **R1** FAIL → window floors outside the txn; reaper-armed TTL; sibling intents
- **R2** FAIL → `park()` bypassed the disarm; sibling branch skipped instead of clearing
- **R3** FAIL → rolling legacy finalize re-armed (root cause: arming was never guarded → structural guard); then my guard broke all claims pre-DDL
- **R4** FAIL → the availability cache never worked in production (`Database.__hash__ is None`)
- **R5** FAIL → reusable-`id()` cache staleness
- **R6** PASS → "Cache, rollout fallback, retention guards, and window-floor behavior are correctly implemented and genuinely tested."

Three of those were regressions introduced by the fixes themselves and caught before merge. Three test-fidelity gaps that had let a fully-green suite hide real defects are also closed: a proxy that intercepted only snapshot reads (not DML), a fake database that is hashable when production is not, and a process-global cache with no test isolation.

Gates: ruff · mypy (190 files) · pytest 2126 passed / 61 skipped.

PR #439, opened 2026-08-04
deferred settlement (2b): a peer plane serves CREDITS traffic without home

jperla:
Peer-plane half of deferred settlement. **Off by default** (`TR_FEDERATION_DEFERRED_SETTLEMENT_ENABLED=false`); the home-side apply endpoint and the forwarder are 2c, so nothing in this PR can move money anywhere yet.

## Why

Today a federated key on the AWS/Azure planes gets 402 `CREDITS_NOT_ON_THIS_PLANE` for every credits-routed request: `PostgresStore` is not a `TypedBillingStore`, so authorize takes the legacy branch and unconditionally calls `STORE.reserve()` against a local balance that federated keys deliberately seed at **zero**. Only BYOK traffic survives a home outage — which caps per-cloud availability at GCP's.

Identity federates; credits don't. Usage is the third category — a **debt record**, safe to record on a peer and apply late, because applying it only moves a home balance *down* and an insert-once claim at home makes double application impossible.

## What

A federated key whose local balance refuses falls back to spending on credit at home's ledger. **Local reserve is tried first** — a workspace explicitly pre-funded by credit transfer must keep spending that transferred money locally rather than stranding it.

Two properties carry the design, and a 3-lens adversarial panel (23 agents) was right that the first draft earned neither:

**The cap is a real bound.** Admission is a conditional `UPDATE ... WHERE outstanding + est <= cap`, rowcount-gated, in the *same transaction* as the authorization insert. A read-then-check bounds nothing — the authorize-to-settle gap is a whole provider stream, so N concurrent requests each read one stale total and each admit. In-transaction also means an idempotent replay can't double-count and a refusal can't orphan an authorization.

**Abandoned work is reclaimed.** The enclave dying between authorize and settle is routine (every deploy), and this plane has no other sweeper — Spanner's `reap_expired_reservations` is Spanner-only. Each abandonment would leak its estimate permanently, and enough of them would 402 a workspace *forever* on a healthy plane. The new reaper claims the same insert-once marker settle claims (first-writer-wins), is outbox-guarded, and refunds the key-limit escrow.

## Verification

Mutation-tested, **6 of 7 caught**: advisory cap, absent reaper, missing outbox guard, missing marker gate, removed counter clamp, and a settle that stops clearing the expiry each fail a test. The seventh (the outbox `ON CONFLICT`) is proven-redundant defense behind the marker gate, which is itself covered — verified by mutating the gate.

Tests run against the SQLite psycopg fake, which executes the store's real SQL; the guarantees live in statement text, so a Python twin would prove nothing. Full suite green (2973 passed), ruff + mypy clean.

One dialect trap worth flagging: `body ->> 'settled'` is the text `'false'` on Postgres and the integer `0` on SQLite. The reaper avoids it entirely — settle clears `expires_at`, and `expires_at IS NOT NULL` is the "still outstanding" predicate.

### DIFF
--- a/src/trusted_router/config.py
+++ b/src/trusted_router/config.py
@@ -443,6 +443,31 @@ class Settings(BaseSettings):
     # presents there. Both empty = this plane cannot send credits anywhere.
     federation_credit_peer_base_url: str = ""
     federation_credit_peer_token: str = ""
+
+    # --- Deferred settlement -----------------------------------------------
+    # Lets a PEER plane serve a federated key's CREDITS traffic while the home
+    # plane is unreachable: the spend is admitted against a transactional
+    # outstanding cap, recorded durably as debt, and forwarded to the home
+    # ledger when home returns. Identity federates and credits do not; usage
+    # is the third category — a debt record, safe to apply late because
+    # applying it only ever moves a home balance DOWN and an insert-once claim
+    # at home makes double application impossible.
+    #
+    # OFF by default. A plane with this unset behaves exactly as before: a
+    # federated key with no local balance gets 402 CREDITS_NOT_ON_THIS_PLANE.
+    federation_deferred_settlement_enabled: bool = False
+    # The bound on how much unsettled debt one workspace may run up on this
+    # plane. Enforced by a conditional UPDATE at authorize, not a read-then-
+    # check — 200 concurrent authorizes must not all see the same stale total.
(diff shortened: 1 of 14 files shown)


jperla:
codex review found 4 P1s; all closed in the follow-up commit and the two money defects are mutation-verified (reintroducing either fails a test). Finding 2's shape also exists in the pre-existing LOCAL settle path on main — spun off as its own task rather than widening this PR.

PR #441, opened 2026-08-04
deferred settlement 2c: home applies recorded debt; the peer forwards it

jperla:
Completes deferred settlement. **Off by default** on every plane; the tokens exist but nothing is enabled. With 2b (#439) merged, this is the half that lets a peer plane's recorded debt reach the home ledger — the last piece before AWS/Azure can serve CREDITS traffic through a GCP outage.

## Home — `POST /v1/internal/federation/apply-usage`

`source_plane` comes from **which token authenticated**, never the body. `TR_FEDERATION_SETTLEMENT_INBOUND_TOKENS` is a `plane=token` map; duplicate token values are rejected, and (per codex review) settings construction now also rejects a settlement token equal to the resolve-key, internal-gateway, or credit tokens — the doctrine that no secret grants both directory reads and money movement, enforced rather than documented.

Insert-once per claim; replays return the **recorded** verdict; debit-only — no path here creates spendable credits. A missing shard-0 balance row on a *live* workspace upserts a recreated zero-credit row (the 2a lesson) rather than dropping the debit; workspace existence is checked against the workspace entity, not the UPDATE's rowcount.

The **aggregate clamp** (per plane, workspace, UTC day) replaces the design-draft per-row ceiling, which bounds nothing when the sender controls row count. Checked before any write, keyed by home's clock.

## Peer — the forwarder

Classification is the safety argument. Dead-lettering fires only on **structured** verdicts (`settlement_terms_conflict`, `workspace_unknown`); a bare 404 is an outage that parks the whole pass, so a home rolled back past this deploy doesn't dead-letter the backlog. Clamped and outage rows back off (`next_attempt_at`), and the drain selects only eligible rows ordered by it — a parked backlog no longer starves fresh rows or hammers a struggling home from every instance's loop. Dead-letter moves `outstanding → dead_lettered`: headroom restored, debt visible. The forwarded decrement rides the conditional state flip and counts only transitions this pass actually made.

Scheduling is in-process (jittered daemon, single-flight, reap every ~10th pass), gated on the peer config existing.

## Verification

codex-reviewed; 4 findings (2 money/security P1, 2 availability) all closed and mutation-verified 5/5. codex separately confirmed the normal retry paths clean — Spanner ABORTED replay idempotent, UTC day stable across replay, crash-after-200 redelivers and decrements exactly once. Full suite green; ruff + mypy clean.


jperla:
codex review found 4 findings (2 money/security P1, 2 availability); all closed in the follow-up commit, mutation-verified 5/5. codex separately confirmed the normal retry paths clean — Spanner ABORTED replay idempotent, UTC day stable across replay, crash-after-200 redelivers and decrements exactly once.

PR #455, opened 2026-08-05
Add Scaleway model provider

jperla:
## What changed

- add Scaleway as a prepaid, EU-focused TrustedRouter provider
- publish all 12 live chat/vision models and both live embedding models
- ingest availability from Scaleway's authenticated catalog
- ingest exact structured EUR pricing and convert it with the ECB daily EUR/USD reference rate using Decimal arithmetic
- add hourly discovery/price refresh, production secret wiring, provider branding, and social assets
- leave Whisper unlisted until TrustedRouter has an audio transcription route

## Safety

- no BYOK, ZDR, confidential-compute, or E2E claim is attached to Scaleway
- live models without an exact structured price fail the refresh closed
- future model IDs are discovered automatically when their provider namespace and price are known

## Validation

- direct live calls passed for 12/12 chat models and 2/2 embedding models
- `uv run ruff check .`
- `uv run mypy`
- `uv run pytest -q` (3066 passed, 223 skipped, 10 xfailed)

The enclave support in Lore-Hex/quill-cloud-proxy must deploy first so these catalog routes are never published ahead of gateway support.


jperla:
Closing as superseded. Scaleway went live in #762 on 2026-08-23 (`a5c34ef1` "Activate IO, Scaleway, Featherless, and Jina providers"). Main carries Scaleway, Featherless and Jina in the catalog with `tests/test_scaleway_featherless_jina.py` covering all three; the pricing-provider approach in this branch was not the one taken.

PR #486, opened 2026-08-08
Pin analytics cutover to the active image digest

jperla:
## Summary
- require every control-plane region to be fully converged on one live release before analytics cutover
- reuse the active revision's resolved image digest instead of inferring an image tag from local Git abbreviation settings
- pass the verified live release identifier into the rollout

## Verification
- bash syntax validation
- 19 focused cutover and operational deploy tests
- ruff clean

This is intentionally left as a draft during the new ClickHouse-primary soak so merging it does not trigger an unnecessary production rollout.


jperla:

PR #524, opened 2026-08-12
Gate the provider-check contract against production drift

jperla:
## Why

A public suite (`Lore-Hex/trustedrouter-provider-check`) will let inference providers test their own endpoint against the attested gateway's wire contract before applying to the marketplace. To predict production, it has to assert what production asserts — so it vendors the PONG prompt, the SSE/response extractors, the rotation error taxonomy, the failure-ownership model, the deadline policy, and the leaderboard scorer.

Vendored copies rot silently. This adds the gate that stops it: an exporter, a committed snapshot, and a CI test that recomputes the snapshot from the live symbols.

## Two pinning mechanisms, because either alone fails open

**Behavioral tables** pin semantics, but only at sampled inputs. An adversarial review mutation-proved the gap: inserting a new `if "minimax" in model_l: return 1024` branch into `_rotation_max_tokens` — a new model family, the likeliest real change — left the snapshot byte-identical and CI green.

**Normalized source hashes** of all 24 vendored functions close that. Normalization strips docstrings, blank lines, full-line comments and indentation, so comment churn stays quiet while any semantic edit trips the gate. Hashes alone would only say "something changed"; the behavioral tables say what the contract *is*, which is what the public repo replays against its copies.

## Coverage holes the review found (all mutation-proved green before this PR)

| Symbol | Hole |
| --- | --- |
| `classify_provider_failure` | no case reached 402, 503, 529, `overload`/`capacity` kinds, bare `status="unsupported"`, or the in-branch 401/403 disjunct |
| `_rotation_max_tokens` / `_rotation_omits_temperature` | grid never crossed a provider gate with a model marker, so deleting `provider_l == "openai" and` changed no row |
| `model_deadlines` | no case exceeded the slow-marker floor, so flattening `max(first_token, 45.0)` was invisible; slow-vs-fast precedence and the `ValueError` guard unpinned |
| `aggregate_leaderboard` | exactly one rank-eligible model and provider, so the comparator was unobservable and reversing the sort passed; the exporter also re-sorted the output, discarding production ordering |
| extractors | `_pong_matches`, `_chat_text`, `_responses_text`, `_sse_line_*`, `_first_int`, `_response_error` were absent entirely |
| `_rotation_error_type` | never called; its inline literals and precedence unpinned |

Now: 74 extractor rows, 113 classification cases, 26 grid rows, 18 deadline cases, and a leaderboard fixture that records the aggregator's own ordering instead of sorting it away.

The catalog branch of `model_deadlines` is pinned through an injected stub module, and the test asserts `trusted_router.catalog` never lands in `sys.modules` — the branch stays cold while its clamps stay pinned.

## Production change

One: `PONG_PROMPT` is hoisted out of three call sites so parity can import it rather than scrape the literal. Pure refactor — the comment recording why the phrasing is load-bearing (the 2026-06-02 rephrase caused 100% `pong_mismatch`) moves with it.

## Verification

Mutation table — 9 must be red, 1 must be green:

| Mutation | Result |
| --- | --- |
| Add a new `minimax` branch | RED |
| Delete the OpenAI provider gate | RED |
| Remove 503 from the capacity set | RED |
| Delete the 402 leg | RED |
| Make PONG matching exact | RED |
| Accept a new SSE delta key | RED |
| Reverse the leaderboard availability sort | RED |
| Flatten the slow-marker `max()` | RED |
| Alter a `_rotation_error_type` literal | RED |
| Add a pure comment | GREEN |

I independently re-ran the three load-bearing ones (new branch, exact-match PONG, comment-only) rather than trusting the agent's table; all three matched.

Guards against a gate that pins nothing: every marker list non-empty, floors on every section's row count, all 24 hashes 64-char hex, none equal to the empty digest, and all 24 distinct (a normalizer collapsing every function to the same non-empty text would otherwise pass).

`build_contract()` is deterministic across runs and re-export is byte-identical. `ruff`, `ruff format --check`, `mypy` (236 files) clean. The parity test plus the loopback-binding suites (`test_gateway_reuse_probe`, `test_per_enclave_probes`) pass unsandboxed: 65 passed.

## Sequencing

[#523](https://github.com/Lore-Hex/quill-router/pull/523) removes a dead leg from `classify_provider_failure`, which is hashed here. If it lands first this branch needs a rebase and one re-export — the gate doing its job. I'll handle that before merging this.

### DIFF
--- a/scripts/export_provider_check_contract.py
+++ b/scripts/export_provider_check_contract.py
@@ -1,0 +1,1017 @@
+#!/usr/bin/env python3
+"""Export the production-derived provider-check contract snapshot."""
+
+from __future__ import annotations
+
+import ast
+import dataclasses
+import hashlib
+import importlib.util
+import inspect
+import json
+import sys
+import textwrap
+from pathlib import Path
+from types import ModuleType, SimpleNamespace
+from typing import Any
+
+import httpx
+
+ROOT = Path(__file__).resolve().parents[1]
+OUTPUT_PATH = ROOT / "src/trusted_router/data/provider_check_contract.json"
+
+
+class _CatalogFreeProvider(str):
+    """Keep leaderboard deadline fallback coverage out of the catalog branch."""
+
+    def __bool__(self) -> bool:
+        return False
+
+
+def _load_script_module(path: Path, module_name: str) -> ModuleType:
+    spec = importlib.util.spec_from_file_location(module_name, path)
+    if spec is None or spec.loader is None:
+        raise RuntimeError(f"could not load {path}")
+    module = importlib.util.module_from_spec(spec)
+    root_text = str(ROOT)
+    inserted_root = root_text not in sys.path
+    if inserted_root:
+        sys.path.insert(0, root_text)
+    try:
+        spec.loader.exec_module(module)
+    finally:
+        if inserted_root:
+            sys.path.remove(root_text)
+    return module
+
+
+def _normalized_source(fn: Any) -> str:
+    """Return stable source text while ignoring docstrings and comment-only churn."""
+    source = textwrap.dedent(inspect.getsource(fn))
+    parsed = ast.parse(source)
+    definition = parsed.body[0]
+    ignored_lines: set[int] = set()
+    if isinstance(definition, (ast.AsyncFunctionDef, ast.FunctionDef)) and definition.body:
+        first_statement = definition.body[0]
+        if (
+            isinstance(first_statement, ast.Expr)
+            and isinstance(first_statement.value, ast.Constant)
+            and isinstance(first_statement.value.value, str)
+        ):
+            ignored_lines.update(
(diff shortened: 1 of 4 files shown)


jperla:
Rebased onto `main` after #523 merged, and the gate proved itself in the process: with no change to the exporter, the parity test went **red** on the rebase because #523 edited `classify_provider_failure`. That is the intended behavior — a production contract symbol cannot move without CI noticing.

Re-exported. The diff between the pre- and post-rebase snapshots is exactly two entries:

- `source_hashes["provider_reliability.classify_provider_failure"]`
- `contract_version` (derived from it)

Every behavioral row is byte-identical — all 113 classification cases, 74 extractor rows, 26 grid rows, 18 deadline cases, and the leaderboard fixture. That is mechanical corroboration of #523's claim that removing the unreachable leg changed no behavior for any input the classifier actually sees: had it altered even one attribution, a behavioral section would have moved too.

It also demonstrates the split working as designed — the source hash caught the edit, and the behavioral tables proved the edit was semantically inert.

PR #525, opened 2026-08-12
Close three holes in the provider-check contract gate

jperla:
## The defect

`test_optional_public_provider_check_snapshot` (added in #524) compares production's snapshot against `<repo>/contract_snapshot.json`. The public package keeps its copy at `src/tr_provider_check/data/contract_snapshot.json` — it is loaded as package data at runtime, and a root-level second copy would just be another thing that can drift. No root-level file exists, so **the gate could never have passed**.

It looked green because it never ran:

- No workflow in this repo sets `PROVIDER_CHECK_REPO_PATH`, so CI always took the `pytest.skip`.
- Setting it by hand raised `FileNotFoundError` rather than comparing anything.

This is the *only* mechanism that detects the public suite's vendored snapshot drifting from production — its own failure message is what tells a maintainer to open a sync PR — and it guarded nothing from the moment it was introduced. My fault: I reviewed #524 by mutating production symbols, which exercised the two tests that do run, and never set the env var to check that the third one worked.

## Fix

Point it at the real path, and assert the file exists instead of letting `read_text` raise — an operator who sets the variable is asking for the comparison to happen, so a wrong path should say so.

## Verified against a real checkout

| Scenario | Before | After |
| --- | --- | --- |
| `PROVIDER_CHECK_REPO_PATH` unset | skip | skip (unchanged) |
| pointed at the real package | `FileNotFoundError` | **3 passed** (was 2 passed + 1 skipped) |
| pointed at a bogus path | `FileNotFoundError` traceback | clear assertion naming the expected path |
| pointed at a deliberately **drifted** snapshot | never reached | fails with `PARITY_MESSAGE` |

That last row is the one that matters: the gate now demonstrably catches divergence rather than merely being present.

`ruff`, `ruff format --check`, `mypy` (236 files) clean.

## Follow-up

Arming this in CI needs the public repo to exist and be cloned by the workflow. That lands with the repo — until then the gate is run manually, which is now actually possible.


jperla:
Added a second commit: `probes._elapsed_ms_with_clock` and `leaderboard._aggregate_providers` were vendored by the public package but pinned by nothing — absent from the source-hash map and from every behavioral table.

`_elapsed_ms_with_clock` is the one that matters. It is the timing primitive behind `_observe_provider_stream`, which *is* hashed — so the caller's fidelity check stayed green while every TTFT and elapsed-ms figure derived from it could drift freely. Two mutations left the public suite fully passing: changing it to report seconds as milliseconds, and deleting the function outright.

`_aggregate_providers` is 68 lines reached through `aggregate_leaderboard`, not distinguished by the exported fixture.

The map goes 24 → 26 entries, still all distinct. Verified after re-export: the seconds-as-milliseconds mutation now turns the public suite red, and reverting restores green.

Both holes came from the same review that found the path bug — worth noting the pattern, since all three are the same failure shape: a guard that is present, looks green, and checks nothing.

PR #571, opened 2026-08-15
Design doc: AWS/Azure execution steps, and correct the failover claim

jperla:
Two things: a correction to something I asserted in the previous revision, and the actual commands for the remaining clouds.

## Correction

I previously argued that a GCP-written v2 envelope can never reach another cloud's enclave, because the databases are separate. **That is true of routing and false in general.**

The Azure enclaves are deployed with:

```
TR_CONTROL_PLANE_BASE_URL="https://azure.trustedrouter.com/v1,https://trustedrouter.com/v1"
```

and `internal/trustedrouter/client.go` fails over to the next endpoint on a dial failure. So if the Azure control plane is undialable, an Azure enclave asks the **GCP** control plane to authorize — and a GCP authorization carries a GCP-database envelope, which is now v2.

The exposure is narrow: `api-azure` is not in the canonical `api.trustedrouter.com` DNS pool, so a caller has to target it explicitly, *and* the Azure control plane has to be down. But that is exactly the conjunction that shows up during an incident — the Azure control plane fails, failover works as designed, and BYOK breaks on top of the outage it was meant to survive.

So step 1 on the Azure enclaves is a prerequisite for that failover path being safe, not only for Azure's own step 2.

## Current state, verified

Both the AWS and Azure control planes **404 on a blog post that merged before the step-2 commit**, so both are still writing v1. Nothing is broken today.

## §4.6 — the actual commands

Neither deploy is runnable from repo access alone, and the doc now says why:

**Azure** needs the `az` CLI and has an ordering trap — the CCE policy hash is a function of the entire container-group definition, and the Key Vault release policy must be widened *before* the new group is created, or the enclave attests correctly, is refused by Key Vault, and exits. Both regions need doing; `api-azure-sea` is a separate container group. The image already contains v2 read support, so this is a redeploy of current `main`, not a code change.

**AWS**'s `deploy.yml` does **not** roll out to instances — it builds, signs trust files, and syncs a static trust page to S3. The enclave update is host-side (`build-eif` → capture PCR0 → `make deploy-trust`). Dispatching the workflow alone publishes trust artifacts for a build that is not running, which is the "stale trust artifacts" failure its own header warns about.

**Verification** for either cloud: check the *running* enclave attests a build containing quill-cloud-proxy#162, rather than trusting a workflow exit code. Both planes currently publish `source_commit: NOT_CONFIGURED`, so `TRUST_AWS_SOURCE_COMMIT` / `TRUST_AZURE_SOURCE_COMMIT` need wiring during this work or there will be nothing to verify against.

No code changes.

### DIFF
--- a/docs/design/byok-aad-v2-migration.md
+++ b/docs/design/byok-aad-v2-migration.md
@@ -8,7 +8,9 @@ whole difficulty — **and it repeats per cloud deployment**, see §4.0.
 
 **Progress:** steps 1 and 2 are done **for the GCP deployment only**
 (quill-cloud-proxy#162, quill-router#560). AWS and Azure have not been
-sequenced. Steps 3 and 4 are not started anywhere.
+sequenced — verified 2026-08-15: both control planes 404 on a blog post that
+merged *before* the step-2 commit, so both are still writing v1. Steps 3 and 4
+are not started anywhere. Execution steps for the remaining clouds: §4.6.
 
 ---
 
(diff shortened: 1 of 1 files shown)


jperla:
Closing — the substance landed in a different form, and the rest describes work that is now done.

**The correction is in main.** This PR argues that "a GCP-written v2 envelope can never reach another cloud's enclave, because the databases are separate" is true of routing but false in general. `docs/design/byok-aad-v2-migration.md` now makes exactly that point (lines ~517-521):

> Per-cloud: rolling back step 2 on one deployment does not affect what the others *write*, since the databases are separate. It does not follow that the read side is per-cloud. Once **any** cloud's control plane has written even one v2 envelope, **every** enclave that can be served by that control plane needs v2 read support permanently.

Same conclusion, stated generally rather than through the specific Azure→GCP `TR_CONTROL_PLANE_BASE_URL` failover path.

**The progress section conflicts and main's is newer.** This branch says AWS and Azure "have not been sequenced" and are "still writing v1" as of 2026-08-15. Main's table now records all three clouds complete through step 3, citing #573 (backfill) and #576 (operator config and explicit-KMS gate) — both merged after this branch was cut.

**§4.6 covers work main reports as finished.** "Executing the remaining clouds" is the sequence for AWS/Azure steps 1-3, which the table above marks complete.

If the concrete failover mechanism is worth spelling out — that Azure enclaves carry a two-endpoint base URL and fail over to the GCP control plane on dial failure, so BYOK can break during exactly the outage failover exists to survive — that is worth adding to the current doc as a short section. It would be a few paragraphs against today's text rather than a rebase of a branch whose progress claims are now contradicted.

PR #575, opened 2026-08-15
Make synthetic fanout test deterministic

jperla:
## Root cause
The post-merge CI runner completed the concurrency test in 186.9ms against a hard 180ms wall-clock threshold. The identical PR run and local full suite passed, so this was runner load rather than a product regression.

## Fix
Replace the timing assertion with direct peak-concurrency instrumentation. This proves probes overlap while removing dependence on runner speed.

## Validation
- targeted xdist test passed
- full synthetic monitoring file: 110 passed
- ruff clean

### DIFF
--- a/tests/test_synthetic_monitoring.py
+++ b/tests/test_synthetic_monitoring.py
@@ -2260,6 +2260,8 @@ async def test_run_synthetic_once_fans_out_targets_and_probes(monkeypatch: pytes
             "https://trusted-router-control-eu.example",
         ),
     ]
+    active_probes = 0
+    peak_active_probes = 0
 
     def fake_probe(probe_type: str) -> Any:
         async def run(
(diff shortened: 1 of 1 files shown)


jperla:
Superseded by #633 (merged as `ba2abb20`), which fixes the same test for the same root cause with a strictly stronger proof, so closing this to avoid a conflicting re-application of the weaker version.

Why #633 is stronger, verified by mutation rather than claimed: an adversarial pass ran three mutations of `probes.py` against both versions. Serializing the target gather and serializing the within-target gather fail under both. But **partial** serialization — targets concurrent, probes chunked 2-at-a-time, so at most 6 of the 10 unmetered probes are ever in flight — **passed the old wall-clock test in 0.46 s** and fails #633. #633 uses a mutual rendezvous (every unmetered probe registers, then awaits an `asyncio.Event` that is only set once all 10 have arrived, under `wait_for(timeout=5)`), so a serialized implementation parks deterministically instead of merely being slow, and `peak_in_flight == 10` becomes a biconditional for full fan-out. It also excludes the six credit-bearing probes on purpose: `DEFAULT_SYNTHETIC_BILLING_CONCURRENCY = 2` caps them, so enrolling them in a ten-way barrier would deadlock by design; their bound stays pinned by `test_run_synthetic_once_bounds_credit_bearing_probe_concurrency`.

Worth noting this PR was open for two days while the same threshold flaked again at 180.2 ms on CI run 31983353289 — the fix landing is what mattered, and the diagnosis here was right.

PR #587, opened 2026-08-15
Gate control-plane deploys on BYOK envelope format ordering (report-only)

jperla:
> **Draft: lands last.** Conflicts with #585 on the migration doc and with #586 on four
> trust files. Merge #583, #584, #585, #586 and Lore-Hex/quill-cloud-proxy#177 first, then
> I rebase this.

## What this enforces

The migration plan's §4.0 warns, in its own words, that the step-2 change arrives on other
deployments as an ordinary version bump rather than a deliberate migration step —
*"Nobody has to decide to run it."* If a control plane writes a format its cloud's enclave
cannot read, every BYOK key there breaks at the next inference request.

Today nothing enforces the order. On AWS and Azure it was not enforced; it was harmless
only because those databases held no BYOK rows.

The invariant: **a control plane may only deploy a build whose written envelope formats are
a subset of the formats accepted by every enclave serving that cloud.** Wired into
`deploy.yml` and both hand-run cloud scripts, since the hand-run path is the one that
skipped the ordering.

## It ships REPORT-ONLY, deliberately

`DEFAULT_MODE = REPORT_ONLY`. It prints the full table and a loud WOULD-BLOCK verdict, and
exits 0.

Landing this enforcing would hard-block every deploy on all three clouds for a reason
nobody can clear from this repository — AWS and Azure have never published a
`source_commit`, let alone one carrying an accepted-formats declaration. That is the same
mistake in kind as the ordering trap this gate exists to prevent: shipping the enforcing
half before the half it depends on exists. It is also stacked on an already-red pipeline
(the GCP control-plane deploy has been failing since 07:50 UTC on an unrelated IAP
permission in the step #574 added).

**The flip is one line** — `DEFAULT_MODE = ENFORCING` — and arms `deploy.yml`,
`aws_eu_control_plane.sh` and `azure_control_plane.sh` together. The documented precondition
is that `--mode enforcing --cloud gcp --cloud aws --cloud azure` exits 0.

Review found an earlier revision claimed report-only "stops nothing" while still exiting 1
on an underivable write side — reachable by ordinary refactoring, on a non-skippable
dependency. Report-only now has no non-zero exit path; `main()` catches `Exception` so a bug
in the gate cannot stop a deploy either.

## Both sides are derived by running code, not by reading it

- **Accepts**: from the enclave's `accepted_formats.json`, generated by a Go test that
  round-trips each format (quill-cloud-proxy#177).
- **Writes**: by installing a recorder on `EncryptedSecretEnvelope.__init__`, calling every
  write entry point for real, and reading `.algorithm` off what those calls produced.
  Spelling is irrelevant to it. Demonstrated on a V3 write built in a module *outside*
  `src/trusted_router` — the AST scan's own named blind spot — which the behavioural probe
  catches and the scan does not.

The AST scan still runs and the gate uses the **union**, which is fail-closed: a larger
written set can only make `written ⊆ accepted` harder to satisfy.

## What it does NOT establish

- A run-time kill switch in the enclave defeats the accepted-formats declaration entirely.
  Nothing here closes it; see quill-cloud-proxy#177.
- Report-only's "exits 0" covers every path `main()` returns from. It does not cover a
  `BaseException` out of a probed import, or an argparse usage error exiting 2. Neither is
  reachable by ordinary refactoring; both are named in the module docstring, `deploy.yml`
  and both scripts.
- The `.algorithm`-assignment refusal is deliberately over-broad — narrowing it means
  deciding which objects are envelopes, which is the inference this scan exists because it
  cannot make. In report-only that costs nothing; it is a real cost once flipped.


jperla:
Superseded by #606. Review concluded the deploy-gate delivery was wrong: three insertion points on the critical path, per-deploy cost forever, and a report-only mode needed solely to make it safe there. #606 keeps the invariant and both behavioural derivations, checks the diff instead of the deploy, and is silent on every PR that does not change the written format set. The quill-cloud-proxy half (#177, the enclave accepted-formats declaration and source_commit publication) is unchanged and still wanted.

PR #600, opened 2026-08-16
Attribute purchases per user and gate phone verification on a first funding

jperla:
Phase 2 of the user-provided Custom Models plan (follows #596).

## Lifetime top-up attribution
Every real purchase now accrues to the **initiating user** — Stripe checkout + ACH (metadata), PayPal (`custom_id` as JSON; the legacy pipe-delimited reference is still parsed so in-flight orders capture correctly), x402 (intent metadata) — by passing `lifetime_topup_user_id` into the existing grant, so accrual rides the event claim and is exactly-once per payment (verified: a webhook replay accrues once; the property fails under mutation). Auto-refill and legacy events fall back to the workspace owner. The signup grant never accrues. `grant_credit.py` does not accrue unless `--count-toward-lifetime-topup` — support grants must not unlock identity verification. `scripts/set_lifetime_topup.py` is the idempotent support override.

Checkout accepts `purpose=identity_verification` and answers with `verification_topup_remaining` as a hint; the $1 minimum is untouched.

## Funding gates phone verification
Both start entry points (`POST /notify/phone/start`, console form) now require lifetime top-ups > 0 — every call or text we place is backed by a completed payment (kills SMS/voice pumping). Enforced by default outside local/test (`phone_verification_requires_funding` tri-state; test corpus unaffected). API → `403 verification_required` with `missing_requirements` + `verification_url`; console → form disabled with the reason and a link to add credits. Confirm/cancel/remove stay ungated. New `GET /auth/verification-status`; OAuth identity payload gains `phone_verified`.

## Verification
`ruff check`, `mypy`, full `pytest -n 4`: 3,873 passed / 284 skipped / 10 xfailed.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -222,6 +222,16 @@ jobs:
           SPANNER_INSTANCE_ID: trusted-router-nam6
           SPANNER_DATABASE_ID: trusted-router
         run: scripts/deploy/migrate_typed_counters.sh
+      - name: Apply money primitives schema (earnings, movements, lifetime top-ups)
+        # tr_user_lifetime_topup is written INSIDE the credit-grant transaction
+        # by every purchase webhook, so a missing table would abort the grant
+        # and stop payments crediting. This job runs before `deploy`, so the
+        # table exists before any revision that needs it can serve.
+        env:
+          GCP_PROJECT_ID: quill-cloud-proxy
+          SPANNER_INSTANCE_ID: trusted-router-nam6
+          SPANNER_DATABASE_ID: trusted-router
+        run: scripts/deploy/migrate_money_primitives.sh
       - name: Apply bounded request-retention schema
         env:
           TR_GCP_PROJECT_ID: quill-cloud-proxy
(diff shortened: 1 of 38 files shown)


jperla:
Adversarial review (3 finder lenses × 2 skeptics per finding; 10 raw → 5 confirmed → 3 distinct issues) — all fixed in 84f64ef8:

- **Would-have-been payment outage:** `tr_user_lifetime_topup` is written inside the credit-grant txn but its migration wasn't on the deploy path and was verified absent in prod. Now wired into `migrate-schema` (runs before `deploy`), AND applied to prod after the in-flight rollout finished — `tr_credit_movement`, `tr_earnings_balance`, `tr_user_lifetime_topup` + index all verified present.
- **Adyen** never accrued (initiator-less signed reference) → now accrues to workspace owner, replay-safe, tested.
- **PII leak:** verification-status resolved ownerless management keys to the workspace *owner* and returned their phone/email → 403 per house rule; regression test fails under mutation.

Refuted (kept as-is): refund/chargeback not reversing accrual (by design), support-override sharing the claim namespace (harmless).

Full suite after fixes: 3,875 passed / 284 skipped / 10 xfailed.

PR #645, opened 2026-08-17
docs: client-telemetry — connect timeout is outcome timeout in both examples

jperla:
Contract v1 doc fix, no behavior change.

The §3.2 retry example read `po=transport_error;pc=connect_timeout`, and the §3.4 forwarded-body example mirrored it as `"prev_outcome":"transport_error"`. The reference implementation (trusted-router-py 0.6.0, `_telemetry.py` `on_transport_error`) classifies every `httpx.TimeoutException` — `ConnectTimeout` included — as outcome `timeout`, so the wire value is `po=timeout` (normalised `prev_outcome=timeout`). Per the document's own header rule ("if this document and those modules disagree, the modules win and this document has a bug — fix the document"), this fixes both examples. The §3.2 `po` enum line already lists `transport_error` and `timeout` as distinct outcomes and is untouched.

Found while pinning byte-for-byte parity vectors for the header-only telemetry PRs (trusted-router-js#14, -go#10, -rust#19, -java#17, -swift#12); their engine-path tests assert `po=timeout` for timeout-class previous attempts, matching py. Enclave-side validation of the SDK headers was verified in production logs (falsification probe drew `client_context_dropped reason=x_tr_client_grammar`; the five SDK smoke ids drew none).

Related, tracked separately: trusted-router-py#15 (py emits `a=100` past the grammar range; sibling SDKs suppress — py fix PR incoming).

### DIFF
--- a/docs/client-telemetry.md
+++ b/docs/client-telemetry.md
@@ -71,7 +71,7 @@ s  = 0|1 streaming
 fo = 0|1 candidate index advanced at least once during this logical request
 ```
 - First attempt: `v=1;a=0;s=1`. Retry after a connect timeout on the apex that moved to an alias:
-  `v=1;a=1;po=transport_error;pc=connect_timeout;ph=apex;pm=10012;sm=10530;s=1;fo=1`.
+  `v=1;a=1;po=timeout;pc=connect_timeout;ph=apex;pm=10012;sm=10530;s=1;fo=1`.
 - Static identity (SDK name/version/runtime/OS) rides the existing `User-Agent`, not this header.
 - **Not sent to custom base URLs** (a self-hosted gateway is not TrustedRouter's to measure) and **not sent
   on control-plane calls**. Not sent when telemetry is opted out (§6.4).
@@ -89,7 +89,7 @@ settlement token and are not joinable in ClickHouse).
 ```json
 "gateway_request_id": "rlog_…",
 "client": {"v":1,"source":"tr","sdk":"tr-py","sdk_version":"0.6.0","lang":"python","runtime":"cpython/3.12.1",
-           "os":"macos","arch":"arm64","timeout_ms":120000,"attempt":1,"prev_outcome":"transport_error",
+           "os":"macos","arch":"arm64","timeout_ms":120000,"attempt":1,"prev_outcome":"timeout",
            "prev_error_class":"connect_timeout","prev_host":"apex","prev_elapsed_ms":10012,"since_first_ms":10530,
            "stream":true,"failover_used":true}
 ```


jperla:
## Merge gate: blocked at `ee5d3fb` — the two lines are right, the rest of the document is not

Independent adversarial review (codex-cli, high reasoning effort), asked to **verify the claim against source** rather than review prose, and to sweep the whole document for other module contradictions.

> `MERGE VERDICT: BLOCK - the two-line timeout correction is right, but the document still materially contradicts the authoritative implementations`

### This PR's change is confirmed correct, independently and by live probe

- `_telemetry.py` `on_transport_error` tests `isinstance(exc, httpx.TimeoutException)` → `outcome = "timeout"` **first**; the `body_started`/`stream_broken` branch is the `elif`, so it cannot divert a timeout. `ConnectTimeout.__mro__` = `[ConnectTimeout, TimeoutException, TransportError, RequestError]`.
- `header_value` emits `f"po={previous.outcome}"` verbatim — no remapping downstream.
- Codex reproduced the corrected string exactly with a live probe: `v=1;a=1;po=timeout;pc=connect_timeout;ph=apex;pm=10012;sm=10530;s=1;fo=1`.

The block is entirely about **other** contradictions found in the sweep. They split three ways, and the split is the point — some of these must **not** be "fixed" in the document.

### A. The document is right; `trusted-router-py` has the bug (do NOT edit the doc)

Editing these would enshrine known defects. All three are already tracked upstream.

- **`po=ok`** — py can emit `po=ok` (reproduced: `v=1;a=1;po=ok;...` via `x-should-retry: true` forcing a retry after a sub-400 response). Codex initially called the doc's `po` enum false; but `client_context.py` `CLIENT_PREV_OUTCOMES` — the executable authority this document's own header rule names for the header/settle vocabulary — is **exactly** the five documented values. So py violates both the doc and the control plane. Tracked as Lore-Hex/trusted-router-py#17. *This independently confirms the cross-SDK `po=none` ruling the five SDK PRs implement.*
- **`a=100`** past the documented `0..99` — Lore-Hex/trusted-router-py#15.
- Caller-supplied `x-tr-client` surviving opt-out — Lore-Hex/trusted-router-py#18.

### B. Genuine document bugs — modules win, safe to correct here

Each verified against source:

| § | Document says | Module says |
|---|---|---|
| §5.4 (~L203) | counter `outcome` is `FinalOutcome` | `client_events_schema.py` is `Outcome` (no `exhausted`) |
| §5.3 (~L164) | `method` permits `PUT\|PATCH\|DELETE` | `client_events_schema.py` is `Literal["GET","POST"]` |
| §6.1 (~L226) | py header assembly is at `_client_sync.py` **and** `_requests.py` "(two sites!)" | it is `_set_recorder_header` in `_transport.py`, called from all four drivers |
| §5.2 | host mapping "anything else→custom" | every HTTPS subdomain of `trustedrouter.com` maps to `control` |
| ~L113 | a 400 drops the batch and keeps going | the reporter **disables telemetry for the process** — and L247 of this same document says so correctly |
| ~L131 | `synthetic` "true iff…" | py hardcodes `"synthetic": False` |

The §6.1 row matters beyond bookkeeping: it is the table the five in-flight SDK PRs (js#14, go#10, rust#19, java#17, swift#12) implement their emit points from. No harm resulted — each SDK located its own single emit point — but the reference is wrong for the next implementer.

### C. Needs a contract-owner ruling — not a mechanical correction

- **§5.1 "no timestamps" / "Nothing else"**, yet every batch carries `sent_at_ms` (`_telemetry.py`, and the schema permits it). Either py sends a timestamp the privacy principle forbids, or the principle is overstated. This sits directly against the cosign-signed trust claims, so it is a policy call, not an edit.
- **§5.4 fold ladder (~L217-219)** omits the arbitrary-existing-key fallback present in `_telemetry.py`, which can **misattribute** dimensions rather than only coarsen them. That bears on the "honest numbers" principle in §2.3.

### Deploy impact: none — correcting the premise this PR was reviewed under

`.github/workflows/deploy.yml` is `paths:`-filtered to `src/**`, `scripts/deploy/**`, `frontend/src/**`, `Dockerfile`, `.gcloudignore`, `pyproject.toml`, `uv.lock`, and itself. **`docs/**` is not listed, so merging a docs-only change triggers no Cloud Run rollout.** Only `ci.yml` runs unconditionally on push to `main`; `prod-smoke` and `trust-drift` chain off the deploy's `workflow_run` and stay idle.

### State

Open at `ee5d3fb`, all 7 checks green, `MERGEABLE`/`CLEAN`. The two changed lines are provably correct and could merge on their own merits; the sweep above is the reason a merge gate did not approve the document as a whole. Full codex reasoning (8,615 lines) is retained locally if wanted.

PR #649, opened 2026-08-17
A cloud is not done until rows are observed moving

jperla:
## A cloud is not done until rows are observed moving

From 2026-08-02 to 2026-08-17 the AWS-EU cloud served production traffic with no
analytics pipeline. The drain that moves rows from the DSQL outbox into
ClickHouse had never been installed, 470,897 rows piled up, `activity_generations`
was empty, and nothing reported it for fifteen days — because the only backlog
alarm is emitted BY the drain that was missing.

The proximate cause was not a missing monitor. `scripts/deploy/aws_eu_clickhouse.sh`
ended by PRINTING `Next: apply clickhouse/*.sql, then redeploy tr-eu with ...`
and exiting 0. A human ran it, read the echoes, and stopped. "The script
finished" and "the cloud works" were different things, and nothing anywhere
treated "cloud exists, has no drain" as an incomplete rollout.

This makes them the same thing, for one cloud, executably.

## What it does

`scripts/deploy/verify_cloud_complete.sh <cloud>` asks five questions, in an
order where each is meaningless until the one before it holds:

| # | Stage | Fails when |
|---|---|---|
| a | in the fleet freshness registry | there is no status endpoint for this cloud, so nothing can read its drain lag and the run has nothing to fetch |
| b | `/status.json` carries the `analytics` section | the cloud publishes no answer to the question |
| c | `analytics.available` is `true` | the control plane cannot read its own outbox (**not** the same as an empty one) |
| d | `drain_lag_seconds` under the bound | rows are enqueued and nothing is deleting them — the AWS-EU shape exactly |
| e | the control plane's outbox is not switched OFF | nothing is enqueued at all, so (c) and (d) pass over an empty pipe |

Every judgement is in `src/trusted_router/cloud_rollout_completeness.py`, unit
tested without a network. The shell owns the ordering, one HTTPS GET, and the
exit status. No credentials, so it runs from a laptop.

Stage (a) means REGISTERED, not WATCHED. The check that reads that registry
(`check-analytics-freshness.yml`) ships with `workflow_dispatch` as its only
trigger, deliberately, until every cloud publishes the section — so the stage
says it has an endpoint, not that anyone is looking at it.

Stage (a)'s URL comes from `ANALYTICS_FRESHNESS_FLEET` (landed in #643) and is
never re-typed here. That is load-bearing: AWS's entry is the tr-eu App Runner
control plane that holds the Aurora DSQL connection — the deployment whose drain
was missing — not `aws.trustedrouter.com`, which fronts the Fargate plane. A
gate pointed at the friendlier hostname would have been green throughout.

The cloud LIST is not re-typed either: `declared_clouds()` delegates to
`operational_analytics_fleet.deployed_clouds()`, the union of every table in
this repo that declares a deployment. Add a cloud to any one of them and CI
fails until it is checkable and checked.

## How "this script runs the gate" is established

The first version of that binding was a regex: does the string
`verify_cloud_complete.sh <cloud>` appear in the script's last N lines? A heredoc
body satisfies it. So does a printed instruction. So does a commented-out line.
The check written to stop "printing the step counts as doing the step" was
itself satisfied by printing the step.

So it is behavioural. `tests/test_deploy_script_execution.py` RUNS each bound
script to completion against a stub PATH and asserts what it DID:

1. it called the gate, for its own cloud;
2. with the gate failing, it exits non-zero;
3. it issues no further cloud CLI calls after the gate answered;
4. both of the gate's exit codes come out the far end unchanged.

Three saboteur scripts — a printed instruction, a commented-out call, a
swallowed exit status — are executed alongside them to demonstrate the fix
rather than assert it. `aws_eu_clickhouse_drain_install.sh` is recorded
`NOT_PROVEN` with a written reason: its middle ships a payload over SSM and
reads the drain's journal back, so a stub answering `Status=Success` would make
that an assertion about the stub.

## What changed in this revision

Two doors that were still open, and a pass over every sentence the tooling
prints.

### 1. GCP — the primary cloud — was bound only by a text match

GCP's `rollout.sh` is exempt from the gate for a real reason: it is a step of the
deploy job, and fetching `trustedrouter.com/status.json` in the middle of
deploying the cloud that *serves* `trustedrouter.com` would abort the deploy that
repairs an outage. But the compensating control was twenty lines of YAML in that
job, and all that bound it was a substring search over the job's `run:` blocks
for `verify_cloud_complete.sh gcp`.

A reviewer replaced the entire job body with

```
echo "Next: bash scripts/deploy/verify_cloud_complete.sh gcp"
exit 0
```

and the suite stayed green at 76 passed. So did the commented-out variant, and
so did `|| true`. All three of the saboteur shapes this PR exists to kill
satisfied the one binding covering the primary cloud: the exception had become
the hole.

The job's body is now **`scripts/deploy/verify_gcp_complete.sh`**, bound in
`ROLLOUT_REGISTRY` and executed by the behavioural harness exactly like the
other four — it calls the gate for `gcp`, cannot exit 0 over a failing gate,
provisions nothing afterwards, and passes both exit codes through. The workflow
step is one line that invokes it, and CI requires a step whose `run` **is** an
invocation of a script this cloud proves by execution; an echo of the command is
not one. Verified by running each saboteur: the echo now fails the workflow
binding, `|| true` fails two behavioural cases, the commented-out call fails
four.

What stayed a declaration, because a workflow cannot be executed from a test:
that the job exists, that it depends on `deploy`, and that it runs
`if: always()`. Said in the exemption text and in the docs' limits list.

### 2. `NOT_PROVEN` was an escape hatch that cost 121 characters

Flipping `aws_eu_clickhouse.sh` from `PROVEN_BY_EXECUTION` to `NOT_PROVEN` with
121 characters of filler took the suite from 76 passed to 71 passed, **green**:
five parametrised behavioural cases stopped being collected, and a test that is
not collected cannot fail. The doc guard passed because it asked whether the
script's basename appeared *anywhere* on the page — which it does, in the prose
explaining why. The registry could say `NOT_PROVEN` while the docs said proven.

The gate is no longer the reason's length. The proven and not-proven lists in
`docs/storage-portability/multi-cloud-separation.md` are parsed out of marked
blocks and compared to `ROLLOUT_REGISTRY` as **sets**, exactly, in both
directions. A script cannot change proof status without an edit to a page
somebody reviews, and the failure names which script moved and which way.

### 3. Every printed sentence is now true

* Stage (a) told an operator on every pass that "somebody reads this cloud's
  drain lag on a schedule". Nothing does. It says what it establishes.
* The corrected harness-hermeticity wording landed last round in
  `tests/deploy_script_harness.py` and the old stronger claim survived in four
  other files, including the module this PR is named for. They agree now:
  isolation is **by name**, bounded by `STUBBED_COMMANDS`.
* The "adding a cloud" checklist prescribed the exact bash 3.2
  heredoc-in-`$( )` parse bomb this PR removed from
  `aws_eu_clickhouse_drain_install.sh`. Fixed, with a note that the new `bash -n`
  test covers this repo's scripts, not one a future author writes from the page.
* "Runs after every production mutation" was false for `verify-cloud-complete`:
  `migrate-schema` and `sync-runtime-secrets` mutate production *before*
  `deploy`, and the job is skipped when `deploy` is skipped. The stated coverage
  is now every run in which the deploy job ran, whatever its result.
* The verifier wrote `__pycache__` into `src/` on the `.venv` branch — a
  read-only check writing into the repository it reads.
  `PYTHONDONTWRITEBYTECODE`.
* The five-stage summary still described the deleted exemption;
  `ScriptExemption.reason` claimed it was "Printed by audit" when nothing prints
  it; the VERIFIED banner said flatly that the control-plane script "enables the
  outbox" when for AWS it may set it `false`; `CloudRollout.deploy_scripts`
  documented a rule the code does not enforce and no cloud satisfies. All four
  now say what is actually the case.
* The two dated "measured today" claims are gone — they were already stale, `gcp`
  VERIFIES now — so the docs point at the command instead of at a date.

### 4. Deleted, and fixed

**Dead code:** `rollout_scripts()` (zero references) and `status_url_for()`
(dead in production — `main()` re-implements it inline — and kept alive by its
own test).

**Two correctness holes,** both of the shape "a check quietly stops being able to
fire": a `generated_at` in the FUTURE makes stage (d)'s age negative, so the
staleness half can never fire again however stale the snapshot is; and stage (c)
tested `available` with Python truthiness, so the string `"false"` read as
available. Both fail now, each with its own sentence.

### Earlier in this PR: it got smaller

Four review rounds each produced findings, landing in the same three places every
time — a verdict-kind taxonomy, seven exit codes, an exemption mechanism. Two of
the findings were regressions introduced by the fixes for the earlier ones. That
complexity was deleted rather than patched again: the eight verdict kinds and the
`TR_VERDICT` sentinel; exit codes 2, 4, 6, 7; the `--max-lag-seconds` /
`--status-url` flags and the DIAGNOSTIC outcome that existed to make them safe;
`analytics_absent_reason`, `apply_exemption()`, `structurally_absent()`,
`drain_lag_caveat()`; `CLOUD_COMPLETE_GATE_DIR`; and the tests that existed only
to police those.

**The contract:** a stage held if its process exited 0. Nothing is parsed out of
a stream, so no interpreter warning can change an outcome. Stdout carries plain
note lines that the shell reprints verbatim under the outcome; they inform and
never re-rank a run.

**Outcomes:** `VERIFIED` (0), `NOT YET OBSERVABLE` (5), `NOT VERIFIED` (1, for
everything else, with the reason printed). 5 is the one extra code and it earns
its keep: it is the state every cloud is in until a control plane that publishes
the section is deployed, and the run that INSTALLS a drain hits it by
construction. Every bound script is proven to pass both codes through.

**There is no way to excuse a stage.** No waiver, no exemption, no flag, no
environment variable. A cloud that cannot be checked is NOT VERIFIED and the run
exits non-zero. `rollout.sh` keeps its exemption and its compensating-control
note as documentation of which FILE ends in the gate; the gate itself waives
nothing, and GCP is now bound like everyone else.

## Claims that were false, and are now either true or gone

* `aws_eu_clickhouse_drain_install.sh` **did not parse under bash 3.2** —
  `/bin/bash` on every macOS. Its next-steps text was `"$(cat <<'NEXT' ... )"`,
  and a heredoc nested in a command substitution is a syntax error there as soon
  as the body contains an apostrophe; the body said "step 9's output". The
  script whose absence caused the outage could not be run by the person most
  likely to run it, while CI on Linux stayed green. Fixed, every deploy script is
  now parsed with the local shell by a test, and the checklist that prescribed
  the same form is fixed too.
* `aws_eu_north_clickhouse.sh` overwrote the gate's exit status with its own 3
  whenever `TR_STOCKHOLM_REPLICA_WIRED` was unset — every first run. The
  propagation claim held only because the harness fixture supplies that
  variable. The gate's answer wins now, and a test asks the question the way a
  first run asks it.
* The `verify-cloud-complete` job claimed to run "after every production
  mutation" on a bare `needs: [deploy]`. It runs on `if: always()` now, and the
  coverage it does have is stated exactly rather than rounded up.
* The runbook said the bound scripts "exiting 0 means the cloud works". It does
  not: it means the check passed, and the check cannot see rows move.
* The harness called itself hermetic while symlinking all of `/bin` and
  `/usr/bin` into its PATH, and the verifier claimed the one mktemp file was the
  only thing it wrote anywhere.
* Stage (e) reported a control-plane script that DOES NOT EXIST as one that
  "never sets `TR_OPERATIONAL_ANALYTICS_OUTBOX_ENABLED`".

## What this enforces, and what it does not

**Enforces.** A cloud in any deployment table with no freshness endpoint or no
rollout entry fails CI. A cloud whose deploy scripts are unbound, or bound
without a reason, fails CI. A script under `scripts/` that runs the verifier
with no registry entry claiming it fails CI. A bound script — for every cloud,
GCP included — that stops calling the gate, or that swallows its status, or that
keeps provisioning after it, fails the execution tests. A script that stops being
executed by the harness fails CI until the docs move with it. A bring-up run
against a cloud with no drain, a stopped drain, a stale or future-stamped
section, an unreadable outbox, or a disabled outbox exits non-zero and names the
fix.

**Does not enforce.** It cannot see rows move — no public status page can, since
an empty outbox and a switched-off one publish the same number; the remaining
evidence is two in-cloud counts ten minutes apart, and the banner says so every
time. Stage (e) reads the working tree, not the deployed revision, and is a text
read that anyone determined can beat. A cloud in none of the deployment tables is
invisible to all of this. One bound script's tail is claimed, not proven. GCP's
job is half declaration: what the check DOES is executed, but that the job exists,
depends on `deploy` and runs `if: always()` is read as text out of the workflow —
and it has never run on a merge; it lands here. `bash -n` covers this repo's
scripts, not one a future author writes from the checklist. And
`ROLLOUT_REGISTRY` does not enumerate a cloud's scripts: it cannot know which of
the sixty files under `scripts/deploy/` belong to which cloud, so a bring-up
script that neither runs the gate nor is named there is caught by a reviewer, not
by CI.

## Where the fleet actually is

Do not read it off this PR — run the gate. At the last reading taken while
writing this, `bash scripts/deploy/verify_cloud_complete.sh gcp` exited 0
(VERIFIED) and `aws` and `azure` both exited 5: their control planes do not
publish the `analytics` section yet, and each starts answering when it is
redeployed. Azure additionally fails stage (e) — it has no operational-analytics
outbox at all.

## Gates

`uv run ruff check .` clean, `uv run mypy` clean (278 files), full suite
**4825 passed / 290 skipped / 10 xfailed** (was 4817 passed: GCP's five
behavioural cases, the docs/registry set-equality test, and two correctness
tests).

Stacked on #643.


jperla:
Superseded by #658 — same diff replayed onto current main. This branch carried #643's commits via a merge, and main now has them squashed, so a rebase replayed already-merged work as 16 conflicting files. #658 is the identical change with a clean history.

PR #652, opened 2026-08-17
Give the AWS analytics nodes their own role instead of the enclave hosts'

jperla:
## What

Both AWS ClickHouse deploy scripts launched their node with `quill-enclave-instance-profile` — the profile the Nitro Enclave hosts use.

That role grants `secretsmanager:GetSecretValue` on `quill/*` and `kms:Decrypt` on `arn:aws:kms:*:330422590279:key/*`. So `tr-eu-clickhouse-1`, an ordinary analytics box with `EnclaveOptions.Enabled=false`, held ~40 provider API keys, the Cloudflare token, the cross-cloud SA key, and could decrypt with **any key in the account** — the CloudTrail CMK included.

## Scope taken from evidence, not from guessing

30 days of CloudTrail, by principal:

| Principal | Actually used |
| --- | --- |
| `i-02e34e58761097671`, `i-0ada95aad6d11aa56` (enclaves) | ~40 provider API keys, `cloudflare-api-token`, `trustedrouter-aws-cross-cloud-sa-key`, ECR pulls |
| `i-025ed1c1766f61290` (`tr-eu-clickhouse-1`) | **one** secret, **one** KMS key, SSM heartbeat |

The new role is that observed set. Enclave hosts are untouched.

## Why this is in the script rather than in a shell history

The split was first applied by CLI against the running instance. That fixed nothing durable — the next run of this script would have handed the node the enclave profile straight back, and a new region would never have had the narrow role at all.

A control that exists only as live cloud state dies with the resource, and can't be evidenced as *operating* across a SOC 2 observation period for anything rebuilt during it. So the script creates what it needs, and Paris also converges the profile on an already-existing instance rather than fixing only the launch path.

## Two things that look wrong and aren't

**`StringEqualsIfExists` is kept deliberately.** EC2 instance-profile assumption doesn't populate `aws:SourceAccount`, so hard `StringEquals` fails **closed** — and not at deploy time, but when cached IMDS credentials next expire, surfacing as an unrelated outage hours later. Every other role in this account uses hard `StringEquals` because every other one is a service principal (`ecs-tasks`, `apprunner`, `events`, `cloudtrail`) that *does* populate it. `quill-enclave-role` is the only EC2-principal role in the account.

**The KMS key is resolved, not hardcoded.** The secret uses the AWS-managed `aws/secretsmanager` key, whose id differs per account and region; a literal ARN would silently break the first time this ran anywhere else.

## Tests

Five new tests, each mutation-verified — reinstating the enclave profile, widening the secret scope to `quill/*`, dropping the role creation, and flipping to hard `StringEquals` each fail exactly one test.

`ruff check`, `ruff format --check`, `mypy` clean; 51 tests pass.

## Live state

`tr-eu-clickhouse-1` is already on the new role (swapped 2026-08-17, verified error-free in CloudTrail). This PR makes that reproducible. The Stockholm node does not exist yet, so its change is forward-looking only.


jperla:

PR #660, opened 2026-08-17
axiom: ship logs from a listener thread, not the request thread

jperla:
First console-latency fix. Cloud Run measurements for authenticated `/console` 200s (2026-08-14..16) showed p50/p90 of 84/539 ms in `us-central1`, 398/629 ms in `us-east4`, and 1540/4074 ms in `europe-west4`.

## Defect and performance fix

`axiom_py.logging.AxiomHandler.emit()` can synchronously flush from the caller whenever its one-second interval has elapsed. A cold POST to `api.axiom.co` measured 551 ms, so an ordinary request that logged could pay an observability-network round trip.

Records now cross a bounded 10,000-record queue to a `QueueListener` shipper thread. Enqueue is nonblocking; under backpressure, observability records are deliberately dropped with a throttled constant breadcrumb rather than delaying a request. The handler-install guard recognizes the queue handler, preventing duplicate queues/listeners on repeated initialization.

## Security boundary

The original draft put the scrubber on the queue handler as a normal filter. Review found that design unsafe: stdlib queue preparation subsequently formatted raw `exc_info` and `stack_info` into `msg`; top-level/underscore extras were incomplete; and the filter mutated the shared caller record.

The final design instead owns `_DroppingQueueHandler.prepare()`:

- shallow-copy the caller record before any formatting;
- scrub the ordinary message and every outgoing key/value recursively, including top-level sensitive keys, underscore extras, e-mail values, nested mappings/tuples, and secret field names;
- remove `exc_info`, cached `exc_text`, and `stack_info` before stdlib formatting (Sentry remains the exception-detail tier);
- scrub the exact prepared `msg`/`message` again after custom formatting;
- if any pathological value or formatter makes preparation fail, drop the record and emit only constant metadata—never `QueueHandler.handleError(original_record)`.

No unsanitized record enters the queue, the real `axiom-py` serializer receives only the sanitized copy, and sibling stdout/Sentry handlers retain the untouched original `LogRecord`.

## Shutdown correctness

Shutdown is idempotent across supported Python 3.11–3.14 behavior. The owner callback stops/drains the listener first, flushes the newly populated Axiom buffer second, and cancels the replacement timer created while draining. This fixes axiom-py's FIFO callback order, where its earlier flush cannot see records that are still queued.

## Red and adversarial proof

- On the vulnerable head, the new queue-boundary regression exposed raw API-key/e-mail canaries from exceptions and stacks.
- Before the widened fix, top-level e-mail/authorization/underscore extras and arbitrary exception narratives also survived.
- Pathological `Mapping.items`, mapping-key `__str__`, and string-subclass normalization failures reproduced raw original `msg`/`args` on stderr through stdlib `handleError`; the final regression now proves queue-empty, one deliberate drop, constant stderr, and no raw data.
- Three independent mutations—restoring inherited prepare, omitting exception clearing, and omitting the post-format scrub—each turn their regression red.
- An independently run 800-record blocked-shipper stress accounted for every record as shipped or deliberately dropped, with no raw canary at the sink.
- A real `axiom-py` FIFO shutdown stress delivered 250/250 unique records, emptied queue and buffer, and canceled the replacement timer.

## Validation

Final rebase is conflict-free on `3314f370` (#669), candidate head `fbc1e064`. #669 changes only SMS templates plus `test_sms_form_preview.py`; that intervening test is included in the exact-head affected run.

```text
exact-head Axiom/property + #669 SMS tests: 72 passed
full suite on the immediately preceding, code-disjoint base: 4914 passed, 296 skipped, 10 xfailed, 0 failed (11m39s)
ruff check .: passed
mypy: passed (280 source files)
changed-file ruff format --check: passed
git diff --check: passed
```

Fresh independent review approved the preceding exact candidate; another exact-head review and replacement GitHub CI are gating the final rebase. No merge will use checks from an older SHA.

The repository-wide `ruff format --check .` remains baseline-red on hundreds of untouched files; no unrelated tree-wide formatting was applied, per `AGENTS.md`.

### DIFF
--- a/src/trusted_router/axiom_config.py
+++ b/src/trusted_router/axiom_config.py
@@ -30,6 +30,7 @@ Design choices:
 from __future__ import annotations
 
 import atexit
+import copy
 import importlib
 import logging
 import logging.handlers
@@ -37,6 +38,7 @@ import os
 import queue
 import re
 import sys
+import threading
 import time
 from collections.abc import Callable
 from typing import Any
@@ -45,7 +47,7 @@ from urllib.parse import urlparse
 from urllib3.util.retry import Retry
 
 from trusted_router.config import Settings
-from trusted_router.sentry_config import _scrub, _scrub_string
+from trusted_router.sentry_config import _is_sensitive_key, _scrub, _scrub_string
 
 log = logging.getLogger(__name__)
 HTTPAdapter: Any = importlib.import_module("requests.adapters").HTTPAdapter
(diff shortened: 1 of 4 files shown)


jperla:
Exact-head review result for `4793deb411460631acd9e6e9fe09134ab94cf421`: **REQUEST CHANGES / do not merge**.

The two originally failing tests are repaired correctly, but adversarial review found a security defect in the feature itself. `_AxiomScrubFilter` runs before `QueueHandler.emit()`, then inherited `QueueHandler.prepare()` formats still-unsanitized `exc_info` / `stack_info` into the copied queued record's `msg`. This violates the PR's central guarantee that no raw PII reaches the queue.

Independent reproduction against this exact head:

```text
queued message:
request failed
Traceback (most recent call last):
  ...
ValueError: credential sk-tr-v1-EXCEPTIONCANARY for ada@example.com

raw_key_in_queue= True
raw_email_in_queue= True
```

The current pre-queue canary test covers an `extra` dict but not exception or stack text. The feature needs a prepared-record scrub (for example, an appropriate `_DroppingQueueHandler.prepare()` override) plus exception and stack canary regressions that are verified red without the fix.

Per the task's explicit stop rule—if the feature is actually wrong, report it rather than weakening tests—I am leaving this unmerged.

jperla:
Final merge gate: exact head fbc1e064 was independently approved after rebasing onto main 3314f370; all seven replacement CI jobs passed (lint, frontend, test, test-post-cutover, browser, BYOK ordering, proofs). Local full suite was 4,914 passed / 296 skipped / 10 expected xfails with zero failures. Squash-merged as f04938d93f26724db1fab9c20d17ba53cdbccb22. The original QueueHandler PII blocker is resolved by copy-before-format, exception/stack removal, full payload scrubbing, fail-closed prepare errors, and drain-then-flush idempotent shutdown.

PR #674, opened 2026-08-18
Give Azure a real analytics pipeline: two nodes, one drain, flag last

jperla:
Azure enqueued nothing and drained nothing. `azure_control_plane.sh` set no
`TR_OPERATIONAL_ANALYTICS_OUTBOX_ENABLED`, so the flag fell back to `config.py`'s
`False` — off by omission rather than by decision — and stage (e) of the
completeness gate failed for exactly that reason.

This builds the cloud-local pipeline, in the order that makes the flag safe.

## What is here

**`scripts/deploy/azure_clickhouse.sh`** — the nodes, one region per invocation
(`uaenorth` first, then `southeastasia`; the order is enforced, not suggested).
uaenorth joins `vnet-prod`, because the drain must reach `tr-azure-pg` over its
private endpoint. southeastasia gets its own resource group, VNet, NSG and Key
Vault, joined by **global VNet peering created from both ends and asserted
`Connected`** — a one-sided peering drops every packet while both resources look
healthy. Nodes bind their PRIVATE address; the subnet NSG admits 8123/9000 from
one CIDR and admits the internet to nothing, including no SSH rule anywhere
(shells are `az vm run-command`). `clickhouse/006` + `clickhouse/009` are applied
to each node and the table **count** is asserted, not the exit status.

**`scripts/deploy/azure_clickhouse_drain_install.sh`** — ONE drain, on the
uaenorth node, fanning out to southeastasia so both copies are written before any
outbox row is deleted. Two drains against one outbox would each DELETE rows the
other had not yet written. It refuses to install until it has measured three
things:

* the private DNS zone for the Postgres private endpoint is linked to `vnet-prod`
  — without it the drain resolves the PUBLIC address of a server with zero
  firewall rules and times out on every connection while the unit sits `active`;
* the second node exists (a silent single-target drain logs `copies=1`,
  indistinguishable from a deliberate one-node deployment);
* the login it was given can `SELECT` **and `DELETE`** the outbox and **cannot**
  read `tr_entities`. A role without DELETE is the quietest failure available:
  every row delivered, every metric healthy, outbox growing forever.

**`scripts/deploy/sql/azure_operational_outbox_drain_role.sql`** — the scoped
role. Its password is generated once into Key Vault by the provisioning script,
which never reads it back, and reaches `psql` through a **pipe** — not argv, not
a file. The nodes fetch their own secrets over IMDS; no secret passes through an
operator's terminal at any stage.

**`azure_control_plane.sh`** — the analytics conditional, mirroring
`aws_eu_control_plane.sh`: the flag is now STATED and computed from whether a
ClickHouse node and its vault secret exist. Also fixes the `APP` default from
`tr-azure` to **`tr-azure-vnet`**, which is what production runs; the file called
itself the source of truth while pointing at a different container app.

## Two regions is durability, not residency

Both nodes hold the **same** rows. This is the Paris+Stockholm pattern: the drain
writes both and deletes only when both accepted, so the failure mode is "the
outbox grows", never "data is lost". No row leaves Azure, and nothing here
replicates to or from GCP or AWS.

## Gate compliance

Both new scripts are bound in `ROLLOUT_REGISTRY['azure']`, both are executed end
to end by `tests/test_deploy_script_execution.py`, and the docs'
`PROVEN_BY_EXECUTION` block names them. **What that proves is control flow** —
the gate is called for `azure`, and every gate exit code comes out unaltered —
and both the registry comment and the docs say plainly what it does not prove:
the remote channel is a fixture under stubs, so the middle asserts against the
stub. Evidence that rows moved is still two counts, ten minutes apart, from
inside the VNet.

`expects_outbox` for azure flips to `True` **in this commit**, which means the
fleet check FAILS for Azure until the runbook has actually been walked. That is
the correct reading of a pipeline that exists in the repository and not yet on
the wire, and the failure should be red for hours, not weeks.

## The runbook

`docs/storage-portability/azure-analytics-runbook.md`: nodes → peering → scoped
role → drain → watch it idle → **then** the flag → redeploy → verify from
outside → start watching. Each stage says what "working" looks like and what to
do when it is not.

**Nothing in it has been executed.** Azure ARM auth was dead on the machine that
wrote this (`AADSTS50132`, an expired session — the cached profile answers
`az account show` from disk while every ARM call fails), so no resource named
here has been observed to exist and three defaults are inferences the scripts
CHECK rather than trust: the free `/24` in `vnet-prod`, the private DNS zone, and
the `Standard_E2s_v5` SKU in southeastasia. The runbook says so at the top and
lists the readings to take first.

## Cost

≈ $270/month: two `Standard_E2s_v5` ($113.15/mo each at uaenorth prices read live
2026-08-18), two 128 GiB Premium SSDs ($21.50 each), two Standard public IPs
(~$3.65 each, egress-only — nothing listens on them), plus $0.25/GB of peering
traffic, both sides billed, carrying one copy of each drained batch.

## Gates

`uv run ruff check .` clean; `uv run mypy` clean (280 files); full suite
**4925 passed, 296 skipped, 10 xfailed** in 7m57s. Every deploy script parses
under `/bin/bash` 3.2 (macOS), which is why no next-steps text in this PR is
built as a heredoc inside `$( )`.

### DIFF
--- a/clickhouse/tr-clickhouse-operational-ingest-postgres.service
+++ b/clickhouse/tr-clickhouse-operational-ingest-postgres.service
@@ -35,6 +35,12 @@ Environment=HOME=/var/lib/tr-clickhouse-ingest
 # connection, and on plain Postgres libpq reads PGPASSWORD from this file
 # rather than from argv, where every local user could read it.
 #
+# AZURE uses that second path. Flexible Server has no IAM token, so the Azure
+# node's file omits TR_POSTGRES_IAM_AUTH entirely and supplies PGPASSWORD for a
+# role granted SELECT, DELETE on the outbox table and nothing else. The HOME
+# line above matters MORE there, not less: libpq looks for the client
+# certificate on every connect regardless of how the password arrives.
+#
 # SECOND DURABLE COPY (optional, additive). Setting
 # TR_OPERATIONAL_ANALYTICS_CLICKHOUSE_REPLICA_HOST turns the drain into a
 # dual-write: it writes each batch to the local node AND to that host, and
--- a/docs/clickhouse-reliability.md
+++ b/docs/clickhouse-reliability.md
@@ -285,10 +285,11 @@ header spells out.
 
 Two states are neither pass nor fail, and are printed as `(unchecked)` on every
 run rather than skipped: a cloud with no public status page (`reason=`), and a
-cloud that legitimately runs no outbox (`expects_outbox=False` — Azure today,
-whose deploy script sets no `TR_OPERATIONAL_ANALYTICS_OUTBOX_ENABLED`). The
-second one becomes a **failure** the day that cloud publishes a real lag, which
-is the day it needs watching.
+cloud that legitimately runs no outbox (`expects_outbox=False`). The second one
+becomes a **failure** the day that cloud publishes a real lag, which is the day
+it needs watching. Azure was its only user until 2026-08-18 and is not any more:
+it now has two ClickHouse nodes, a drain, and a control-plane script that states
+the flag, so all three clouds are measured and none is excused.
 
 To ask about one cloud during an incident:
 
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -1134,8 +1134,11 @@ failure mode this exists to stop.
 starts answering when a control plane built from the publisher is deployed to
 it, and exits 5 until then. The last reading taken while editing this section
 was `gcp` VERIFIED with `aws` and `azure` both at 5, and it is a note about a
-moment rather than a claim about now. Azure additionally fails stage (e): it has
-no operational-analytics outbox at all. See
+moment rather than a claim about now. Azure used to fail stage (e) as well —
+its control-plane script set no outbox variable at all — and now passes it
+statically, which means only that the script CAN enable the outbox, not that any
+deployed revision has. Walking it from nodes to flag is
+`docs/storage-portability/azure-analytics-runbook.md`; see
 `docs/storage-portability/multi-cloud-separation.md` §7 for the full definition
 of done and the checklist for a new cloud.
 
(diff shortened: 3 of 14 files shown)


jperla:
Superseded — both halves have landed or are in flight, cut against current main instead of this branch.

**The flag is in [#805](https://github.com/Lore-Hex/quill-router/pull/805), merged.** That was this PR's real contribution: `azure_control_plane.sh` set no `TR_OPERATIONAL_ANALYTICS_*` variable, so the outbox flag fell back to `config.py`'s `False` — off by omission rather than by decision. Azure has had a working ClickHouse node and drain in production since 2026-08-22, so redeploying from a clean checkout would have pushed the outbox **off** over a running pipeline.

#805 adds one thing this branch didn't: failing to *look* cannot disable analytics. Discovery runs through the `az` CLI, and `outbox=false` on a failed lookup would let an expired login or a renamed resource group stop a live pipeline while the deploy reported success. It refuses instead, before any mutation, and derives `expects_outbox` from the fleet registry so script and registry cannot drift.

**The ClickHouse provisioning is in [#719](https://github.com/Lore-Hex/quill-router/pull/719).** Only 6% of this branch's `azure_clickhouse.sh` survives verbatim there — they are independent implementations of the same file, not an evolution, so only one could land. #719 is the newer, post-incident one and has since had six review findings fixed, including missing migrations 010–013, a credential leak through `set -eux` into cloud-init logs, and a non-idempotent Key Vault rotation that would break a re-run against the live node.

**Why not rebase this instead.** It predates main's refactor and still adds `TR_INTERNAL_GATEWAY_TOKEN` and the `TR_FEDERATION_*` settlement variables that main **retired** (its own `RETIRED_OBSERVER_ENV_VARS` list), and still carries the pre-VNet `APP_ENV=$STACK-env` that main fixed to `$STACK-env-vnet`. Merging it would reintroduce retired configuration into a live control plane. #805 pins all five retired names as absent.

Two pieces here are genuinely unclaimed and worth re-cutting if you want them: `scripts/deploy/sql/azure_operational_outbox_drain_role.sql` (a least-privilege drain role) and `docs/storage-portability/azure-analytics-runbook.md`. Neither is in #719 or #805.

PR #694, opened 2026-08-20
Ship the CSP report-only first, then enforce

jperla:
Follow-up to #691, which merged enforcing. Flipping to report-only until the two paths I could not exercise are proven by real traffic.

**What was verified** (browser, under an enforcing policy): `/`, `/choose`, `/docs`, `/chat` — zero console violations, including the four jsdelivr libraries and the same-origin iframe.

**What was not:** `/console/*` (needs a session) and the Adyen checkout page, whose third-party SDK injects its own iframes — the most likely thing here to trip a strict policy.

Under enforcement a blocked script doesn't warn; it silently doesn't run, and checkout is a money path. Report-only costs nothing and the flip back is one word.

Unchanged: per-request nonce, 23 nonced inline blocks, the guard test that fails on an unnonced template, and the three baseline headers (unconditional).

### DIFF
--- a/src/trusted_router/middleware.py
+++ b/src/trusted_router/middleware.py
@@ -325,8 +325,15 @@ def register_http_middleware(app: FastAPI, settings: Settings) -> None:
             # It has NO report-uri, so nothing collects these automatically:
             # they appear in the browser console and nowhere else. That is a
             # deliberate first step, not a monitoring claim -- see the PR.
+            # REPORT-ONLY first, deliberately. The nonces are already correct
+            # (verified in a browser under enforcement on /, /choose, /docs and
+            # /chat), but /console/* needs a session and the Adyen checkout page
+            # loads a third-party SDK that injects its own iframes -- neither
+            # was exercised. Report-only lets real traffic prove those paths
+            # before a blocked script can break a money path. Flip this to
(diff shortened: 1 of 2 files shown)


jperla:
Merging. Checked the live enforcing policy on prod and found a concrete gap that supports this PR's reasoning rather than contradicting it:

```
script-src  ... https://checkoutshopper-live.cdn.adyen.com   ok
frame-src   ... https://checkoutshopper-live.cdn.adyen.com   ok
connect-src 'self' api/trust/status.trustedrouter.com        no Adyen
```

Adyen's Drop-in loads its library from the `.cdn.` host but makes its runtime XHR calls to `checkoutshopper-live.adyen.com` — **no `.cdn.`**. That origin appears in neither `connect-src` nor `frame-src`, so under enforcement those calls are blocked on a live payment path.

Static checks cannot catch this. Every inline block on every reachable page carries the header nonce (verified on `/`, `/models`, `/providers`; `/console/*` is 302 behind a session), and the template guard covers authed templates too — but Adyen injects iframes and issues requests at runtime, which no template scan sees.

Report-only will name the exact origins in violation reports instead of guessing at them, which is the point of doing it in this order.

PR #712, opened 2026-08-21
Add six-service rollout, recovery, and cutover (stack 4/4)

jperla:
Six-surface split, part 4/4: journaled rollout, recovery, and cutover. Authored by codex; reviewed, rebased, and hardened by Fable. Base: `fable/six-surface-edge-iam`. **Draft until the production cutover this PR describes has been executed and verified.**

## What it does
- **Journaled, resumable state machine** for the initial split: every stage transition (staged → reconcile_intent → reconciled, etc.) is CAS-committed to durable GCS state before the mutation it describes; a killed deploy resumes exactly where it stopped, and drifted postconditions fail closed.
- **Atomic cutover**: the six-backend URL map is rendered, provider-validated, and imported as one operation. The legacy monolith is untouched and stays warm; rollback restores the prior URL map only.
- **Recovery bundle**: authority object + per-epoch release bundle + promotion journal under a single conditional IAM binding, so a fresh operator can resume or roll back a dead deploy from durable state alone.
- Mandatory authenticated browser smoke (Playwright) gating every promote phase.

## Fable's commits on top
1. **CI wiring** (`a7abf8a2`): `TR_CONSOLE_SERVICE` corrected to `trusted-router-console` (the workflow shipped the legacy monolith name, which rollout.sh rejects — first post-merge deploy would have failed); frontend-attestation capture step added; runbook provisioning commands for the durable-state bucket.
2. **Reconcile gate** (`e0ed5a5d`): rollout.sh hard-failed whenever `TR_DEPLOY_RECONCILE_LB != 1`, but the workflow deliberately passes 0 for secondary regions (reconcile-once on the shared global LB) — three of four regions would have failed every production deploy. Read-only LB verification now runs unconditionally; only mutation is gated; skip is legal only when classification found nothing inactive. Verified red-without/green-with. Also: recovery provisioning contract corrected on three counts codex confirmed (state URI inside the bundle, exactly one IAM binding, epoch assigned as manifest sha256).
3. **Mechanism pins** (`d2438a81`): seven tests updated from superseded mechanisms to the rewrite's equivalents (provider secret arrays, pinned observer-secret version, bootstrap read-only coverage) — each contract verified preserved before any test changed.

## Not deployment-evidence yet
Cloud Armor route throttles ship preview-only; provider-side restricted Stripe/SES key permissions cannot be attested from Secret Manager; the cutover runbook (docs/runbooks/ddos-edge-hardening.md) is the operator contract.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -401,19 +401,22 @@ jobs:
       TR_WEBHOOKS_SERVICE: trusted-router-webhooks
       TR_INTERNAL_SERVICE: trusted-router-billing
       TR_BILLING_SERVICE: trusted-router-billing
-      TR_ROLLOUT_MANIFEST: ${{ runner.temp }}/tr-rollout/manifest.json
       # The frontend identities are a reviewed contract (like the service
       # inventory above): capture verifies the live LB against them and
       # rollout.sh refuses to run without the resulting artifact.
       TR_FORWARDING_RULE: trusted-router-control-https
       TR_HTTPS_PROXY: trusted-router-control-https-proxy
-      TR_ROLLOUT_FRONTEND_ATTESTATION: ${{ runner.temp }}/tr-rollout/frontend-attestation.json
       SHA: ${{ needs.build-image.outputs.sha }}
       IMAGE: ${{ needs.build-image.outputs.image }}
       # Keep the workflow value raw. rollout.sh normalizes preserve/false/true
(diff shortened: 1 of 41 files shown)


jperla:
Superseded by the folded six-surface split PR: https://github.com/Lore-Hex/quill-router/pull/736 — the three stacked PRs were only consistent as a whole (PR2's six-surface test server depends on the URL-map rename that lived in PR4).

PR #719, opened 2026-08-21
Give the Azure cloud a ClickHouse of its own

jperla:
`azure_control_plane.sh` already exits non-zero saying this cloud is incomplete — it publishes `analytics: not_configured` because there is nowhere to drain **to**. This is step 1 of the four it names, the one it deliberately refuses to make on its own because it's a cost decision.

A port of the **shape** of `aws_eu_clickhouse.sh`, not a reuse. What carries over is the ClickHouse config, the schema in `clickhouse/*.sql`, and the rule both clouds follow: each cloud owns its analytics, no cross-cloud replication, operational rows about Azure traffic stay in Azure.

## Four things differ, each because Azure does

- **No VPC connector.** App Runner needed one because its default egress is AWS-managed NAT with dynamic addresses. The Azure control plane is already integrated into `vnet-prod`, so it routes to another subnet directly.
- **Its own subnet.** `snet-aca` is delegated to `Microsoft.App/environments` and cannot host a VM. `10.61.3.0/24` is free.
- **The password goes through Key Vault and is fetched by the node** at first boot with its own managed identity — never through cloud-init, which is readable from inside the VM via IMDS.
- **`D2s_v3`**, matching `m5.large`'s 2 vCPU / 8 GiB. Chosen on quota: the v5 families are limited to **0** in uaenorth, while DSv3/DASv4/BS have 10 each against a regional cap of 18.

## The schema is applied by the node itself

Not left in `NEXT_STEPS`. That gap is exactly what ran for fifteen days on AWS-EU — a node that is up with no tables looks identical to one that is working.

## What it refuses to do

The role assignment the node needs is **checked and not created**. Creating role assignments from a deploy script is how a deploy pipeline quietly becomes an admin; it prints the command and stops.

The check reads **ARM**, not `az role assignment list --assignee`. That form goes through Microsoft Graph, and this tenant has it hanging past 120s with a valid token — and its failures are indistinguishable from "the grant does not exist", so an outage would read as a missing role and abort the deploy by accusing a correctly-configured identity.

## Tests

Ten static guards, each on a property whose removal still produces a node that provisions cleanly and fails only later — as data nobody can read, or an analytics store on the public internet.

They read **executable lines only**. The script names what it refuses to do ("never 0.0.0.0/0", the `az role assignment create` it prints as instructions), so a whole-file substring test cannot tell a prohibition from a violation — it would fail on the prose or, worse, pass on a real regression once someone reworded a comment.

Mutation-verified on the two that matter most: a public IP and Internet-wide ingress each fail their guard.

## Not in this PR

Steps 2–4: wiring the control plane env, installing the drain, and `verify_cloud_complete.sh azure`. **Enabling the outbox without a drain recreates the AWS-EU outage exactly** — the outbox grows and the only process that alarms about the backlog is the one not installed.


jperla:
Rebased onto main (merge `02b86051`) and reviewed. The **rebase is clean** — main already carries the `expects_outbox=True` registry flip (`c5c67b54`), so I took main's side on all four conflicting hunks and dropped this branch's `_OUTBOX_FREE_ENDPOINT` (main's `_outbox_free_fixture()` covers the same three cases). Net delta is now just the two provisioning scripts and their tests. An independent review confirmed that resolution dropped no executable behaviour.

**Not merging yet** — adversarial review returned BLOCK, and I verified the three most serious findings myself against this tree:

1. **The schema is missing migrations 010–013.** `azure_clickhouse.sh:229` concatenates only `006` and `009`. `clickhouse/013_activity_generations_workspace_id_single_node.sql:1` says it in its own header: *"MUST be applied to a node before the clickhouse/ tree is next shipped there: the updated drain inserts the workspace_id column, and an un-migrated node rejects the insert."* The drain does insert it (`ingest_operational_outbox.py:28`). Azure has a **live node draining right now** (`status.json` → `drain_lag_seconds: 8.8`, `available: true`), so re-running the installer against it makes every insert fail on the unknown column while the unit stays `active` — shard failures are contained, so the daemon looks healthy and delivers zero rows.

2. **The ClickHouse password and the managed-identity bearer token leak into cloud-init logs.** The bootstrap sets `set -eux` and then assigns `TOKEN=$(...)` and `CH_PW=$(...)`. xtrace prints expanded assignments, and cloud-init captures stderr to `/var/log/cloud-init-output.log`. The password is also in local argv (`az ... --value "$pw"`, line 152) and remote argv (`clickhouse-client --password`, line 284).

3. **The installer ships the working tree, not `HEAD`.** It tars `-C "$REPO_ROOT" clickhouse src/trusted_router` directly (line 115), bypassing `scripts/deploy/_clickhouse_bundle.sh`, which main added precisely to refuse dirty/untracked source and archive `HEAD`. The other two installers already use it. As written, an operator can deploy uncommitted Python to production.

Also flagged, not independently verified: the Key Vault secret is rotated before the VM-exists check (line 144) while an existing VM skips cloud-init (line 331), so a re-run can leave consumers holding a credential the server never adopted; and the new tests are substring assertions that would stay green through every failure above (e.g. `test_azure_clickhouse_deploy.py:105` pins exactly `006`/`009`, so it cannot catch the missing migration).

The gap this PR closes is real and worth closing — production is running an Azure ClickHouse node whose provisioning exists in no merged branch. But these scripts would break the live node if run against it, so they need the fixes first.

jperla:
All six findings fixed in `1753e326`. I verified the three I'd independently confirmed, plus the one new thing this turned up.

**Migrations are now derived** from `clickhouse/*_single_node.sql` rather than a hardcoded `006`+`009`, with a comment naming why each other file is excluded. The test derives its expected set from the directory too, so a new standalone migration can't be omitted by forgetting this script — which is the failure mode that produced the original bug.

**A pre-existing bug fell out of deriving that set.** `011` and `013` were qualified `tr.`, but every consumer of the single-node schema uses database `default` — AWS's control plane (`CLICKHOUSE_DATABASE:-default`), AWS's drain installer (`CH_DATABASE:-default`), and Azure. Only GCP uses `tr`, and GCP runs the replicated `ON CLUSTER` migrations, not these. So those two `ALTER`/`CREATE` statements were targeting a database no single-node deployment reads. They're now unqualified and consistent with `006`/`009`; both are idempotent (`CREATE TABLE IF NOT EXISTS`, `ADD COLUMN IF NOT EXISTS`), so re-running is safe.

**Credentials.** Cloud-init is `set -eu` — the `x` is gone, so expanded assignments no longer reach `/var/log/cloud-init-output.log`. The password reaches `az` through a mode-0600 file instead of `--value`, and `clickhouse-client` reads `CLICKHOUSE_PASSWORD` from the environment instead of `--password`. Zero remaining argv uses of either.

**Delivery is proven, not assumed.** Preflight now authenticates to both Postgres and ClickHouse rather than checking DNS and TCP. Final verification sums row counts across four tables and polls until the total advances, exiting non-zero if it doesn't.

**Re-running against the live node is safe.** The existing-node branch is decided before any vault, identity, network or VM mutation, so a re-run no longer rotates a credential the server never adopts. Existing nodes are validated on identity, subnet, ClickHouse auth, and schema state.

**Payload comes from `_clickhouse_bundle.sh`**, which refuses dirty or untracked source and archives `HEAD`. All three installers now share one pin.

191 tests pass; `bash -n`, `shellcheck -x`, ruff and mypy clean.

One thing for elsewhere, not this PR: `aws_eu_clickhouse.sh`'s NEXT_STEPS block tells the operator to apply `clickhouse/00*.sql`, which matches 001–009 and silently skips 010–013 — the same defect class, in prose rather than code. AWS-EU is healthy right now (`drain_lag_seconds` ~1.7), so its node evidently has the column; this is a trap for the next node built from those instructions. Filed separately.

PR #726, opened 2026-08-22
Run synthetic inference probes through trusted-router-py so the SDK telemetry path is exercised every pass

jperla:
Routes the synthetic monitor's two inference probes (`openai_sdk_pong`, `responses_pong`) through the official Python SDK (`trusted-router-py` ≥ 0.6.0) instead of raw `httpx`, so every synthetic pass exercises the SDK's client-telemetry path end to end: the per-attempt `x-tr-client` header (enclave → settle context) and the SDK's beacon to `/v1/client-events`.

Owner decision (2026-08-21): "the synthetics should use the SDKs; we're trying to test the SDKs and client analytics." Motivation measured the same day: **no TR-SDK request had ever settled in production** (zero `client_source='tr'` rows) because the probes were raw `httpx` (UA `python-httpx`, no header) and the canary batch is hand-built. Synthetic package only; no analytics, rollup, methodology, status, or `/v1/client-events` route changes.

## Authorship and review

Authored by codex-cli from a written spec; reviewed by Claude (Fable): the new adapter read in full; `ruff` and `mypy` clean on the changed modules; focused suites (`test_synthetic_sdk_probes`, `test_synthetic_monitoring`) **117 passed** under the reviewer's run; reviewer mutation (`telemetry=True` → `False` in the adapter) failed **3 of 4** SDK-probe tests. Author's full-repo gates: `ruff` clean, `mypy` clean (292 files), `pytest` **5 111 passed, 301 skipped, 10 xfailed**; every new test mutation-checked.

## What the monitor measures is preserved

- Exactly one attempt per probe: `max_retries=0`, `regional_failover=False`, `regional_affinity=False` (the target *is* the region); a 503 carrying `x-should-retry: true` still yields one request (tested).
- The monitor's own `httpx.AsyncClient` is injected, keeping its timeouts and TLS context (attested targets trust the TEE-minted certificate via the attestation probe, not a CA); all four 7 s timeout phases are captured on the outbound SDK request (tested).
- The per-region `api-<region>.quillrouter.com/v1` base stays the SDK `base_url`; captured requests carry exactly `x-tr-client: v=1;a=0;s=0` and a `trusted-router-py/…` User-Agent (tested).
- Synthetic marking keeps both server-recognised forms (`metadata.trustedrouter_synthetic` and `app="TrustedRouter Synthetic"`); the beacon is sent with the monitor key, which `/v1/client-events` identifies as synthetic server-side regardless of the batch's client bit.
- Telemetry explicitly on at sample rate 1.0 (overriding env opt-outs — the monitor wants every event); the reporter is flushed after each target's inference pair, **off the event loop** (`asyncio.to_thread`) so the ≤ 2 s join never lands in another probe's latency; close failures are logged and swallowed, the SDK's atexit hook is the backstop.
- `_chat_text` / `_responses_text` (hashed by the provider-check contract) are byte-for-byte unchanged; SDK payloads are wrapped in an in-memory 200 response after the monitor's timer stops.

## Outcome mapping (typed SDK errors → the monitor's existing taxonomy)

| raw-httpx observation | SDK surface | recorded sample |
|---|---|---|
| 200, JSON object with PONG | dict | `up`, 200, `output_match=true` |
| 200 without PONG | dict or `TrustedRouterError(200)` | `down`, 200, `pong_mismatch` |
| 4xx/5xx | typed `TrustedRouterError` | `down`, original status, `pong_mismatch` |
| `httpx.HTTPError` before a response | `InternalError(503)` **caused by** the httpx error (explicit `__cause__` only) | `down`, httpx class name, no status/TTFB |
| 200 non-JSON | bare `ValueError` | `down`, 200, `pong_mismatch` |
| unexpected exception | — | `down`, exception class name (the anomaly shows on the page instead of as a gap) |

One unavoidable timing difference: SDK JSON decoding now happens before the monitor's timer stops (the old code decoded after capturing latency), so measured latency includes the first JSON decode.

## Deploy

`Dockerfile` runs `uv sync --frozen --no-dev`, so the runtime SDK dependency ships in the control-plane image. The deploy workflow's optional-surface detector enables the synthetic-job step for any `src/trusted_router/` change and runs `scripts/deploy/synthetic.sh`, redeploying both `trusted-router-synthetic-*` jobs with the same image — but that step is `continue-on-error: true`, so verify the job redeploy succeeded rather than assuming it. The control plane now depends on its own SDK (`pyproject.toml`, `uv.lock`).

After the first pass on the new image, the decisive check is `activity_generations` rows with `client_source='tr'` and `client_sdk='tr-py'`, and `client_minute_counters` rows with `sdk='tr-py'` from the monitor workspace.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,12 @@ dependencies = [
   "python-multipart>=0.0.20",
   "sentry-sdk[fastapi]>=2.20.0",
   "stripe>=11.0.0",
+  # The synthetic monitor runs its inference probes through the official
+  # Python SDK so every pass exercises the SDK client-telemetry path
+  # (x-tr-client header + /v1/client-events beacon). It lives in the main
+  # dependencies because the Cloud Run jobs share the control-plane image,
+  # which installs with `uv sync --no-dev` and no extras (Dockerfile).
+  "trusted-router-py>=0.6.0",
   "uvicorn[standard]>=0.32.0",
 ]
 
--- a/src/trusted_router/synthetic/cli.py
+++ b/src/trusted_router/synthetic/cli.py
@@ -77,6 +77,8 @@ async def _one_probe_pass(
             monitor_region=monitor_region,
             api_key=api_key,
             billing_semaphore=limiter,
+            # The inference probes' SDK sessions beacon to this plane.
+            control_plane_base_url=control_plane,
         )
     )
     if not api_key:
--- a/src/trusted_router/synthetic/inference_sdk.py
+++ b/src/trusted_router/synthetic/inference_sdk.py
@@ -1,0 +1,151 @@
+"""The monitor's TrustedRouter Python SDK session for its inference probes.
+
+The pong probes (``openai_sdk_pong``, ``responses_pong``) are the monitor's
+only real model calls, and they used to go out as raw ``httpx`` POSTs. That
+left the SDK's client-telemetry path -- the per-attempt ``x-tr-client`` header
+the enclave forwards into settle context, and the beacon the SDK posts to
+``/v1/client-events`` -- with no traffic from our own monitor: not one
+``client_source='tr'`` row had ever settled in production, and the canary
+batch the monitor posts (``probes.client_telemetry_canary_probe``) is
+hand-built, so it proves ingest liveness and nothing about the SDK. Sending
+the probes through the official SDK makes every pass exercise that path end
+to end.
+
+The session is configured to preserve what the probe measures:
+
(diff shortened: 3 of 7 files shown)


jperla:
Regression from this PR, found 10 minutes after its deploy: the new top-level `"app"` body key is outside the enclave's Responses allowlist (`validateResponsesFields` ⇒ 501 `not_supported_in_alpha`), so `responses_pong` has been down/501 since 03:12 UTC 2026-08-22 (chat unaffected — its request type ignores a body-level `app`). Fix: #728. The tr path itself works: first `client_source='tr'` rows at 03:14 UTC.

PR #728, opened 2026-08-22
Drop the `app` body key the gateway's Responses validator rejects (responses_pong 501 since 03:12 UTC)

jperla:
## What broke

Since the #726 deploy landed at **03:12 UTC 2026-08-22**, every `responses_pong` synthetic sample has been **`down` / HTTP 501** (35/35 in the first ten minutes; up/200 for the 140 samples before it). `openai_sdk_pong` on the same SDK session stayed up.

Root cause, verified from both ends:

- #726 added a new top-level key to both pong bodies: `"app": "TrustedRouter Synthetic"`.
- The gateway's Responses validator (`quill-cloud-proxy` `enclave-go/internal/adapter/responses.go`, `validateResponsesFields`) allowlists the top-level keys of a `/v1/responses` body and answers **`501 not_supported_in_alpha`** for any non-null key outside it. `app` is not in the allowlist; `model`, `input`, `max_output_tokens`, `temperature`, `metadata` are.
- The enclave log for one failing probe: `enclave.request_end … route="/v1/responses" status=501 outcome="server_error" body_bytes=180 attribution="validation"`. 180 bytes is exactly the compact JSON size of the responses body **with** `app` (148 without).
- Chat tolerates the key only because its request type declares `App string \`json:"-"\`` — a body-level `app` is silently ignored there (attribution comes from `x-title` / `x-openrouter-title` headers). The key was inert on chat and fatal on responses.

## Fix

Remove `app` from both pong bodies so each is byte-for-byte what the pre-#726 raw-httpx probes sent. The `inference_sdk` docstring no longer advertises the key and says why the bodies are frozen. The test that asserted `body["app"]` now pins each body's **exact top-level key set** (chat: `{model, messages, max_tokens, temperature, metadata}`; responses: `{model, input, max_output_tokens, temperature, metadata}`) with a comment naming the validator, so adding a key to either body fails CI instead of the status page.

Not touched: the two other `"app"` literals in `probes.py` (settlement usage records, a different contract).

## Authorship and review

Authored by codex-cli from a written spec; reviewed by Claude (Fable). Reviewer re-ran the CI gates on the tree (`ruff check .`, `mypy`, `scripts.check_format_ordering`, the two synthetic test files) and an independent mutation — re-adding `"app": "x"` to the responses body — to confirm the new assertion fails on the bug and passes after restore. Results in the PR conversation.

## Deploy note

[...]

PR #731, opened 2026-08-22
Capture the active rollback revision exactly

jperla:
## Incident

A tagged Cloud Run traffic entry appeared first in status.traffic, so the deploy workflow captured an older tag-only revision instead of the actual 100%-traffic revision. The staged rollout removed the verified baseline from traffic and would have rolled back to the wrong revision.

## Fix

- resolve the sole positive 100%-traffic revision from the full Cloud Run service JSON
- ignore tag-only zero-percent entries
- fail closed on split, missing, malformed, duplicate, or unreadable traffic state before any production mutation
- cover the exact tag-first incident and workflow wiring

## Verification

- focused deploy/rollout tests: 24 passed on current main
- full suite on the patch-identical prior base: 5,306 passed, 301 skipped, 10 xfailed
- Ruff: green
- Mypy: 297 files green
- independent P0-P2 review: approved

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -393,14 +393,19 @@ jobs:
       - name: Capture pre-deploy revisions (for rollback)
         id: prev
         run: |
+          set -euo pipefail
           for region in us-central1 europe-west4 us-east4 southamerica-east1; do
-            rev=$(gcloud run services describe "${SERVICE}" \
+            service_json=$(gcloud run services describe "${SERVICE}" \
               --region="$region" --project="${PROJECT_ID}" \
-              --format='value(status.traffic[0].revisionName)' \
-              2>/dev/null || true)
+              --format=json)
+            if ! rev=$(python3 scripts/deploy/resolve_active_revision.py \
+                <<<"${service_json}"); then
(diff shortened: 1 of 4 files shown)


jperla:
Post-merge independent review record (Claude / Fable, with a codex-cli adversarial read-only pass at effort=high on head `7dfc5b5`): **VERDICT: SAFE TO MERGE**. Bug confirmed beyond the capture step: with main's `status.traffic[0].revisionName`, `scripts/deploy/staged_traffic.sh:37` and `:52` would route 90% then 50% of traffic to the tag-first revision and a canary trip would restore it at 100%. The resolver accepts exactly one positive-traffic entry at 100%, ignores tag-only zero-percent entries, and aborts before any mutation on split/missing/malformed state or a failed `gcloud` call; the same output feeds staged traffic, the primary canary rollback, the billing rollback and every secondary-region rollback (`deploy.yml:479/489/506/523`). actionlint 1.7.12, YAML parse, `bash -n`, `git diff --check` clean; no `${{ }}` expressions added; 23 focused tests pass locally. One non-blocking note: `int()` would truncate a non-integer `percent`, which Cloud Run's schema cannot produce. Rollout run 32552113626 (main@aa0b4fd) also carries #728, the synthetic `responses_pong` 501 fix.

PR #730, opened 2026-08-22
Split runtime surfaces and make OAuth exchange atomic (stack 2/4)

jperla:
Six-surface split, part 2/4. Authored by codex; reviewed, rebased onto the squash-merged PR1 (#714, now `main`), and gated by Fable. Replaces #711, which GitHub closed when PR1's branch was deleted on merge.

- Splits the control process into **console / chat-proxy / webhooks** surfaces with per-surface body caps and worker budgets, so a chat-proxy flood cannot starve console logins and webhook floods cannot starve either.
- **OAuth code exchange is one transaction**: code redemption, session creation, and cookie issue commit atomically; a crash mid-exchange leaves no half-redeemed code and no orphaned session.
- Per-surface secret sets: a surface's environment carries only the secrets it owns.
- Playwright now runs against a dedicated six-surface test server (`tests/browser/six_surface_server.py`) that builds deterministic settings (rate limiting off, outbound network blocked), which supersedes the harness env flag PR1 carried.

Serial landing: merges only after PR1 is verified in production and this PR's own CI is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #713, opened 2026-08-21
Enforce exact edge and runtime IAM contracts (stack 3/4)

jperla:
Six-surface split, part 3/4: exact edge and runtime IAM contracts. Authored by codex; reviewed, rebased, and gated by Fable. Base: `fable/six-surface-app-oauth`.

- **Cloud Armor as a reviewed artifact**: per-surface security policies (host-deny + per-source ceilings enforced; two route-shaped throttles in preview) are rendered and imported as full-policy replacements — a rule added by hand in the console does not survive reconciliation, and the verifier rejects supersets.
- **Runtime IAM as an exact contract**: six runtime service accounts with per-secret accessor grants and per-role storage grants; the read-only verifier (`rollout_iam_verify.sh`) fails closed on any grant on an unconfigured resource, cross-service-account impersonation, or direct folder/org bindings.
- Per-resource SecretAccessor owner matrix replaces blanket deploy-account access.

Gate: covered by the stack-tip full suite (see PR 4); all 64 IAM-verify tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #732, opened 2026-08-22
Bridge the legacy combined rollout until the service split

jperla:
Fixes the production boot blocker caused by #714 landing before the six-service cutover in draft #712.

- keeps deployed combined mode fail-closed by default
- requires an explicit temporary migration opt-in in both Settings and create_app
- makes rollout.sh refuse before cloud setup unless the guarded workflow opts in
- preserves the pre-#714 gateway and Stripe startup requirements
- documents the exact removal condition when #712 lands

Validated locally: 326 focused tests, full Ruff, mypy (297 files), and Bash syntax are green. Full pytest and an independent P0-P2 review are running on the exact commit.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -353,6 +353,9 @@ jobs:
       # Keep the workflow value raw. rollout.sh normalizes preserve/false/true
       # only after reading the actual 100%-traffic primary revision.
       TR_REGIONAL_QUOTA_LEASE_ISSUANCE_ENABLED: ${{ inputs.regional_quota_lease_issuance }}
+      # Temporary explicit opt-in for the pre-split all-routes service. Remove
+      # with the six-service production cutover in #712.
+      TR_ALLOW_DEPLOYED_COMBINED_SURFACE: "true"
     steps:
       - uses: actions/checkout@v4
         with:
--- a/docs/operations/service-surfaces.md
+++ b/docs/operations/service-surfaces.md
@@ -1,7 +1,10 @@
 # Service-surface bulkheads
 
 The TrustedRouter FastAPI image runs as an explicit, fail-closed process role.
-Production must never use the local/test-only `combined` role.
+Deployed `combined` mode is rejected by default. The sole temporary exception
+is the guarded GCP compatibility bridge described under Integration status;
+it exists only because the #714 enforcement landed before the #712 production
+split and requires a second, explicit opt-in.
 
 | Surface | Internet ownership | Background ownership | Capacity contract |
 |---|---|---|---|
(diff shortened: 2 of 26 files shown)


jperla:
Review from the six-surface stack owner (Fable; PRs #714 merged, #730 / #713 / #712 pending), checked against the split design:

**Compatible with the stack, and necessary.** Confirmed independently that `main` as merged cannot deploy to the combined monolith: `synthetic.sh` refuses observer-token jobs against the legacy service and requires `trustedrouter-observer-internal-token`, which only PR3's `secrets.sh` provisions (never run in prod); and #714 moved synthetic ingestion auth to that token while the live jobs still send the gateway token. #714 should not have merged ahead of that contract without this bridge — that was my error.

On the mechanism:
- `TR_SERVICE_SURFACE=combined` + `TR_ALLOW_DEPLOYED_COMBINED_SURFACE=true` reuses the stack's own surface knob with a value the split never deploys, gated to local/test otherwise. No conflict with #730's console/chat/webhooks surfaces or #712's rollout-mode detection.
- `TR_RATE_LIMIT_ENABLED=false` on the legacy backend is correct and worth stating loudly: the current LB does not overwrite `X-TrustedRouter-Client-IP`, so #714's fail-closed source limiter would either reject or single-bucket all Internet traffic. Net effect: under this bridge production gets #714's body admission, framing strictness, and bounded webhook verification, but **not** per-source rate limiting — that arrives only when #713/#712 install the edge identity. The bridge comment already says #712 removes the exception; #712 will rebase over this and delete the block as its own cleanup.

Nothing blocking from my side. One request: keep the "delete this block when #712 lands" comment verbatim — it is the tripwire that makes the cleanup findable.

jperla:
Review note from the synthetic-monitor side (Claude / Fable, after reading the bridge against main): this makes the *service* boot in combined mode, but `scripts/deploy/synthetic.sh` on main (from #714) still refuses unconditionally — lines 18–24 require `TR_BILLING_SERVICE` ("refusing observer-token jobs against the legacy service", and it must differ from `SERVICE`) and lines 38–39 require the Secret Manager secret `trustedrouter-observer-internal-token`, which `scripts/deploy/secrets.sh` has never created in production. The workflow's "Deploy synthetic monitor Cloud Run Job" step is `continue-on-error: true`, so a bridged deploy will go green while the monitor jobs stay on their current image (`de2d15e`). Consequences: the `responses_pong` 501 fix (#728) never reaches the monitor and the status page keeps false-alarming; and since #714 moved synthetic ingestion auth to the observer token, the un-redeployed jobs' beacons would be rejected by the bridged service unless the bridge also accepts the legacy gateway token for synthetic ingestion while jobs are unredeployed (does it? `config.py`/`main.py` here preserve the gateway startup requirement, but I could not find the ingestion-auth fallback). Suggestion for the same bridge flag: under `TR_ALLOW_DEPLOYED_COMBINED_SURFACE=true`, `synthetic.sh` deploys the jobs against `SERVICE` with the legacy gateway token (pre-#714 behaviour) and skips the observer-token requirement, with a test pinning that the refusal still fires when the flag is absent.

jperla:
Seconding the synthetic.sh gap above (stack owner). Verified on `main`@aa0b4fd: `scripts/deploy/synthetic.sh` lines 18–24 refuse without `TR_BILLING_SERVICE` (which must differ from `SERVICE`) and lines 38–39 require `trustedrouter-observer-internal-token`, which is created only by PR3's `secrets.sh` and does not exist in production. With `continue-on-error: true` on the job step, a bridged deploy goes green while the monitor jobs stay on `de2d15e` — so #728's fix never reaches them and the status page keeps false-alarming.

The bridge's `synthetic/internal_auth.py` + `routes/internal/_shared.py` changes (accepting the legacy gateway token for ingestion under `TR_SERVICE_SURFACE=combined`) keep the un-redeployed jobs' beacons flowing, which is right; the missing piece is the job redeploy itself. The minimal addition proposed above — under `TR_ALLOW_DEPLOYED_COMBINED_SURFACE=true`, deploy the existing jobs unchanged against `SERVICE` with the gateway token, with a test pinning the refusal when the flag is absent — is consistent with the split: the folded six-surface PR replaces that path with `TR_BILLING_SERVICE` + observer token and deletes the bridge block. Either that, or drop `continue-on-error` so the refusal is loud. Not blocking the bridge itself, but without it the deploy this PR enables leaves the monitoring regression in place.

PR #733, opened 2026-08-22
Prove the full-queue drop structurally instead of by wall clock (flake held deploy gate-on-ci)

jperla:
`tests/test_axiom_nonblocking.py::test_a_full_queue_drops_instead_of_blocking` asserted that fifty `handle()` calls against a full `maxsize=1` queue finish in under 0.5 s. Under `pytest -n 4` on a shared runner that mostly measures fifty `prepare()` scrub passes, not whether the put blocks, and it failed on main at `aa0b4fd` (CI run 32551919490: "a full queue blocked for 0.630s", 1 failed / 6040 passed) — nothing in that commit touched the queue. Because `deploy.yml`'s `gate-on-ci` requires same-commit CI green, the flake held a production deploy for a 25-minute re-run on 2026-08-22 at ~04:57Z.

## Change

The property is structural in `_DroppingQueueHandler.enqueue` (`put_nowait` → `queue.Full` → `_note_drop`), so the test now proves it structurally: the overflow `handle()` runs on a daemon thread; the test asserts the call **returned** (`Event.wait(timeout=5)` — the bound is only a hang guard), that the item was **dropped** (`qsize()` unchanged at 1) and **counted** (`_dropped == 1`), and releases queue capacity in `finally` so a deliberately blocking mutant's thread can be joined rather than leaked. One test file, +19/−6.

## Evidence

- Load loop, whole file, `-n 4`, 10 repetitions: **10/10 pass** (author) and **10/10 pass** (reviewer, independently).
- Mutation — `enqueue` changed to blocking `put`: **FAILED** in ~5 s (`AssertionError: a full queue waited for capacity`), not a hang; restored: pass (author: `1 failed in 5.29s` / `1 passed`; reviewer: fails at 8 s wall / `10 passed in 0.70s`).
- `ruff check .` clean; `mypy` clean (297 files); `git diff --check` clean.

## Authorship and review

Authored by codex-cli from a written spec; reviewed by Claude (Fable), who applied one hardening (`suppress(queue.Empty)` around the `finally` release so an unexpectedly empty queue cannot mask the assertion above it) and re-ran every check above on the final tree.

**Merge note:** hold until the #732 bridge deploy has passed `confirm-current-main` — a new main commit before that resets the bridge's same-commit CI gate.

### DIFF
--- a/tests/test_axiom_nonblocking.py
+++ b/tests/test_axiom_nonblocking.py
@@ -37,6 +37,7 @@ import queue
 import threading
 import time
 from collections.abc import Iterator, Mapping
+from contextlib import suppress
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


jperla:
Reviewed (Fable). Approve. The proof is now structural: one overflow `handle()` on a daemon thread must return (the 5 s `Event.wait` is a hang guard, not a timing assertion), the queue must stay at size 1 (item dropped), and the drop must be counted; the `finally` drains capacity so a blocking mutant's thread joins instead of leaking, and `not caller.is_alive()` closes the loop. Reducing 50 calls to 1 loses nothing — the 50 only existed to amplify a wall-clock measurement that was the flake. Mutation evidence (blocking put fails in ~5 s) matches what the structure predicts. This removes the one non-deterministic test that failed `main`'s deploy gate tonight.

PR #738, opened 2026-08-22
Say plainly that a remediator decision row is a point-in-time observation

jperla:
`remediator.py` records a sample per (playbook, subject, 30-minute bucket) **only while a condition is present** — `status="down"` means "condition present" — so a condition ending is encoded as **silence**.

That ambiguity has already cost real diagnostic time. On 2026-08-22 the newest row for `heartbeat-stale:job:regional-quota-reconcile` read *"no beat for 20424s (last 2026-08-21T18:55:00.000Z); the scheduler behind this job is dead or wedged"*. It was written at 00:35:24; the condition actually ended at 00:40:00 when the reconciler's heartbeats resumed. Read nine hours later it looked like a live, unreported condition, and produced a bug report against a loop that was behaving correctly.

## Change 1 — explicit resolution markers

When a detector pass finds that a key **this process observed as present** is no longer present, it records exactly one `status="up"` row (`condition cleared after Ns`) and forgets the key, so a recurrence reads as a fresh onset. Properties held deliberately:

- A process that never observed the condition **never** emits an all-clear (no false clears from a freshly started instance).
- A detector that **raises** keeps its conditions active — exception isolation must not manufacture recovery. Enforced by scoping resolution to the detector that owns the key.
- Resolutions do not page; the row is still recorded.
- Recording stays best-effort: a storage failure is logged and the key forgotten, so one clear observation makes at most one attempt.
- `remediation` stays in `OPS_PROBE_TYPES` — never a component, never an SLO, never the freshness clock. `components.py` is untouched.

## Change 2 — honest dedup docs

The docstring promised "one row per (playbook, subject, 30-minute bucket)". `_DECISION_MARKS` is an in-process dict, so with N control-plane instances the store receives up to N rows per bucket (observed 3–7/hour against a documented ~2/hour). Corrected to state the per-process guarantee and to tell readers to deduplicate by `(target, bucket)`. No distributed coordination added — a per-pass store query or lock is not worth it for an ops timeline.

## Evidence

Author (codex-cli) and reviewer (Claude/Fable) both ran the suite; reviewer's independent run on the final tree:

- `tests/test_remediator.py` — **13 passed**; `ruff check .` and `mypy` (297 files) clean.
- Author mutation: resolve unconditionally (ignoring what the process observed) → `test_fresh_process_does_not_resolve_absent_condition` **failed**; restored → passed.
- Reviewer mutation: drop the detector-ownership filter → `test_decision_rows_bucket_and_dedupe` **failed**; restored → 13 passed.
- Reviewer-written throwaway test for the property no listed test covered explicitly — a detector that raises mid-life must emit **no** resolution row and must keep the condition active: **passed** (not shipped; the ownership filter is what makes it hold).

### DIFF
--- a/src/trusted_router/synthetic/remediator.py
+++ b/src/trusted_router/synthetic/remediator.py
@@ -24,10 +24,12 @@ retention, and /fleet rendering path as every other ops signal. A row records
 a condition OBSERVED AT ITS TIMESTAMP, not current state; a condition ending
 is recorded as the ABSENCE of later rows, never an explicit marker, so readers
 must check the underlying signal (for heartbeat-stale, that subject's heartbeat
-rows) before concluding it is still live. Deduplication is process-local: each
-control-plane instance records at most one row per (target, 30-minute bucket),
-so the store can receive one per instance and readers must deduplicate by
-(target, bucket). status="down" means "condition present". remediation rows
+rows) before concluding it is still live. Deduplication is best effort: each
+control-plane instance TRIES to record at most one row per (target, 30-minute
+bucket), but the mark is checked and set without synchronisation, so the
+background loop and the internal endpoint can both write within one process,
+and every instance keeps its own marks. Duplicates are expected; readers must
+deduplicate by (target, bucket) rather than counting rows. status="down" means "condition present". remediation rows
 are in OPS_PROBE_TYPES: never a component, never an SLO, never the freshness
 clock.
 
@@ -64,7 +66,8 @@ from trusted_router.synthetic.alerts import ops_alert
 logger = logging.getLogger(__name__)
 
 REMEDIATION_PROBE = "remediation"
-# At most one decision row per (playbook, subject, bucket) per process.
+# Best effort: one decision row per (playbook, subject, bucket) per process.
+# The check and set are unsynchronised, so duplicates are possible.
 DECISION_BUCKET_SECONDS = 30 * 60
 _DECISION_MARKS: dict[str, int] = {}
 
--- a/src/trusted_router/synthetic/remediator.py
+++ b/src/trusted_router/synthetic/remediator.py
@@ -20,11 +20,14 @@ MODES (settings.remediator_mode):
 
 DECISIONS ARE SAMPLES. Each decision is recorded as a synthetic sample
 (probe_type="remediation", target="<playbook>:<subject>") — the same store,
-retention, and /fleet rendering path as every other ops signal, and one row
-per (playbook, subject, 30-minute bucket) so a persistent condition reads as
-a timeline, not a firehose. status="down" means "condition present" so the
-timeline semantics match every other probe. remediation rows are in
-OPS_PROBE_TYPES: never a component, never an SLO, never the freshness clock.
+retention, and /fleet rendering path as every other ops signal. Each process
+records at most one onset row per (playbook, subject, 30-minute bucket), so
+the store can receive one row per control-plane instance for the same bucket;
+readers counting conditions should deduplicate by (target, bucket). A process
+also records at most one resolution row for each onset it observed.
+status="down" means "condition present" and status="up" means that condition
+cleared. remediation rows are in OPS_PROBE_TYPES: never a component, never an
+SLO, never the freshness clock.
 
 DETECTORS (v1, all T0 blast radius — detection and paging only):
   * stale heartbeats     — a scheduler this deployment expected to beat has
@@ -49,6 +52,7 @@ from __future__ import annotations
 import logging
 import time
 import uuid
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
 
@@ -59,8 +63,10 @@ from trusted_router.synthetic.alerts import ops_alert
 logger = logging.getLogger(__name__)
 
 REMEDIATION_PROBE = "remediation"
-# One decision row per (playbook, subject, bucket): a condition that persists
-# for six hours should read as ~12 rows on a timeline, not 180.
+# At most one onset row per (playbook, subject, bucket) per process. A
+# persistent condition therefore produces about 12 rows per process over six
+# hours, and the store's row count scales with the control-plane instance
+# count. Readers counting conditions should deduplicate by (target, bucket).
 DECISION_BUCKET_SECONDS = 30 * 60
 _DECISION_MARKS: dict[str, int] = {}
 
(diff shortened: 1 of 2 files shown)


jperla:
codex-cli adversarial review **BLOCKED** the first head and was right — the change manufactured the false all-clear it exists to prevent.

**Finding 1 (the serious one).** A detector returning no decisions is ambiguous: "evaluated, condition gone" vs "could not tell". `_detect_monitor_stale` reads the newest 50 samples and filters `OPS_PROBE_TYPES` *after* the read, so once ops rows (the remediator's own 5-minute heartbeat, peer_monitor, client_telemetry_ingest, and the new resolution rows) crowd the last real probe out of that window, a **dead** probe fleet yields an empty result — and the pass wrote `monitor-stale:probe-fleet` **cleared**. I reproduced it independently against the pre-fix code with a throwaway test: `[('monitor-stale:probe-fleet', 'up')]`. Same test passes on the fixed head.

**Fix:** detectors return a `DetectionResult` carrying `authoritative`; only an authoritative pass may resolve its own keys. `_detect_monitor_stale` is non-authoritative when it has no real sample to measure or the timestamp will not parse. A detector that raises is never authoritative.

**Findings 2–4, all accepted:**
- Active entries are now keyed by owning detector + playbook + subject, so two detectors reporting one key cannot clear each other's (impossible today — each detector uses a distinct playbook — but the structure no longer depends on that).
- A lock guards resolve-and-pop: the in-process loop (`main.py:295`) and the HTTP-triggered pass (`routes/internal/synthetic.py:391`) do not share a semaphore and could both write.
- A failed resolution write keeps the key active for the next pass instead of forgetting it, with a deterministic id so the retry cannot double-write. The old behaviour left the durable timeline falsely `down` forever — the exact lie this PR removes — and the old test blessed it.

Evidence on the new head: **18 passed**, ruff and mypy clean; author's mutation (make `_detect_monitor_stale` authoritative when its filtered list is empty) fails `test_ops_rows_crowding_probe_window_do_not_resolve_dead_fleet`; reviewer's independent reproduction fails pre-fix and passes post-fix. Fresh adversarial verdict on this head gates the merge.

jperla:
Second adversarial review **BLOCKED** again — non-authoritative heartbeat/route reads could still false-resolve, a clear/recur overlap could store `down, up` with no onset, and the retry was not idempotent against a store that appends unconditionally.

Rather than chase full distributed correctness for a best-effort ops timeline, I **narrowed the feature**. Both blocked rounds were the same shape of mistake: inferring "the condition ended" from an empty detector result, when empty can equally mean *the detector could not tell*.

Resolution is now an explicit per-detector **opt-in**, carried as the set of durable keys **positively observed healthy in that exact pass** (default: empty). Only `heartbeat-stale` opts in, and only for a subject whose row says `stale is False` within a non-empty row set it actually evaluated.

- A subject that merely **vanishes** from the rows → no resolution.
- An **empty** read or one that **raises** → no resolution.
- `monitor-stale` and `route-quarantine` → **never** resolve; their emptiness is absence of evidence, not evidence of health.

This keeps the case that motivated the PR (a wedged scheduler whose heartbeat returns — the one that misled me this morning) and drops the inferences that cannot be made safely. Retained from the earlier rounds: per-detector keying, the lock around resolve-and-pop, bounded retry, OPS_PROBE_TYPES exclusion.

Evidence: **25 passed**, ruff and mypy clean. Author mutation (let `monitor-stale` opt into resolution) → `test_monitor_stale_never_resolves_even_with_fresh_positive_sample` failed; restored → passed. Reviewer's own three throwaway tests on this head: absent-but-unproven subject is **not** resolved; empty and raising reads resolve nothing; a proven-healthy subject **does** resolve — all passed.

jperla:
**Approach abandoned and replaced.** Three adversarial rounds each found a real defect in recording an explicit resolution row, every one the same shape — inferring that a condition ended when the evidence could not support it:

1. an empty detector result treated as "resolved", so a **dead probe fleet** whose evidence was crowded out of a bounded 50-row read produced a false all-clear (I reproduced this independently against that head);
2. after narrowing, non-authoritative heartbeat/route reads could still false-resolve, and a clear/recur overlap could store `down, up` with no onset;
3. after narrowing to positive per-subject health evidence, a race remained: a healthy pass snapshots, a stale **recurrence** completes, the delayed pass then publishes `up` and drops the active condition — storing `down, up` while the condition is live again.

That last one is precisely the lie the feature existed to remove, so a fourth attempt is not warranted for an ops-timeline nicety. All resolution machinery is gone; this branch is now reset to `origin/main` plus a small, safe change.

**What ships instead.** The original problem was human: a reader (me) took a row's text as *current state* when it described the moment it was written. So `recent_decisions()` now reports each row's `observed_age_seconds` and marks it `point_in_time` — additive, existing keys untouched, so `/fleet` cannot imply currency. The module docstring states plainly that a row records a condition **observed at its timestamp**, that a condition ending is the **absence of later rows** and never an explicit marker, and that a reader must check the underlying signal (for `heartbeat-stale`, that subject's heartbeat rows) before concluding it is still live.

It also carries the dedup correction that was always uncontroversial: the docstring claimed one row per (playbook, subject, bucket), but the marks are process-local, so the store receives up to one per bucket **per control-plane instance** (observed 3–7/hour against a documented ~2/hour). Readers deduplicate by (target, bucket). No distributed coordination added.

Net: 21 lines in `remediator.py`, no new state, no locks, no store writes, no change to detection or recording. **9 passed**, ruff and mypy clean. Reviewer note: `_age_seconds` returns `None` for unparseable or empty timestamps and raises only on `None`, which `created_at: str = field(default_factory=iso_now)` makes unreachable — and the `/fleet` call site is already exception-guarded.

jperla:
Fourth block, and a fair one: the honesty fix was itself still overstating. `_DECISION_MARKS` is checked and set **without synchronisation**, so duplicates are possible *within* a process (the background loop at `main.py:299` and the internal endpoint at `routes/internal/synthetic.py:392` can both write), not merely across instances.

Corrected wording — no code change, since the racy check-and-set is pre-existing behaviour this PR does not touch and is harmless for an ops timeline: deduplication is **best effort**; each instance *tries* for one row per (target, 30-minute bucket); duplicates are expected; readers must deduplicate by (target, bucket) rather than counting rows.

That is now the third iteration of getting one docstring to state only what is true, which is itself the point of the change. 9 passed, ruff and mypy clean.

PR #740, opened 2026-08-22
Keep client-observed availability internal (off every public surface)

jperla:
The public status page gates the "Client-observed availability" section on snapshot **freshness** only. With 100% synthetic canary traffic (6,838 events / 1 tenant / **0 real** in 24h) the live page currently renders five window cards each reading *"Availability insufficient data — 0 requests from 0 tenants / Coverage n/a"*, plus a *"24h all traffic — calibration view"* row whose numbers are all zero.

An empty section on a public status page is worse than its absence.

## Change

`client_observed_status_section` returns `available: False, reason: "insufficient_real_data"` when the gated 24h window has fewer than `MIN_REAL_REQUESTS_24H` (= 1) requests, and the template renders **nothing** for that reason — the `<section>` element included, so no empty styled box remains in the DOM.

| snapshot state | page |
|---|---|
| gated 24h has real requests | full section, calibration row included |
| **no real requests yet** | **nothing rendered** |
| stale (`reason == "stale"`) | stale banner still rendered — unchanged |

`stale` means we *had* client data and lost it: a real warning, deliberately kept visible. Methodology, gated-vs-all-traffic semantics, and the authenticated console view (`templates/console/activity.html`) are untouched — an empty block is fine in an operator page. The distinct reason string keeps the two invisible cases apart for ops and tests.

The threshold is one real request, named and commented: gated windows already exclude synthetic traffic, so a single request distinguishes real client use from a calibration-only snapshot. The goal is "hide while empty", not "hide until popular".

## Evidence

Author (codex-cli), reviewer (Claude/Fable). Reviewer ran the **live production snapshot** through the new projection rather than fixtures alone:

```
LIVE production snapshot -> available=False  reason=insufficient_real_data
STALE snapshot           -> available=False  reason=stale
REAL-traffic snapshot    -> available=True   state=calibrating
```

- `tests/test_client_reliability.py tests/test_public_client_reliability.py` — **55 passed**; `ruff check .` and `mypy` (297 files) clean.
- Reviewer mutation A — let zero-request snapshots through as available → `test_status_html_hides_client_section_without_real_traffic` **failed**; restored → 7 passed.
- Reviewer mutation B — gate the template on `available` instead of the reason (which would also hide the stale warning) → `test_status_html_keeps_stale_client_warning` and `test_status_html_contains_client_section_and_handles_no_data` **both failed**, confirming the stale path is pinned.
- Author mutation: allowing zero requests through failed the same hide test.

Reviewer note: while running mutation B I reverted the template with `git checkout`, which also discarded the author's template edit; it was reconstructed and every check above re-run on the final tree.

### DIFF
--- a/src/trusted_router/client_reliability.py
+++ b/src/trusted_router/client_reliability.py
@@ -559,12 +559,16 @@ def client_observed_status_section(
     windows = raw_windows if isinstance(raw_windows, Mapping) else {}
     raw_24h = windows.get("24h")
     window_24h = raw_24h if isinstance(raw_24h, Mapping) else {}
-    if _int(window_24h, "requests") < MIN_REAL_REQUESTS_24H:
+    gated_available = _int(window_24h, "requests") >= MIN_REAL_REQUESTS_24H
+    all_traffic = _status_all_traffic(snapshot.get("all_traffic"))
+    all_traffic_24h = all_traffic["windows"]["24h"] if all_traffic is not None else {}
+    if not gated_available and _int(all_traffic_24h, "requests") < 1:
         return {"available": False, "reason": "insufficient_real_data"}
     generated_at = snapshot.get("generated_at")
     return {
         "available": True,
-        "state": "published" if published else "calibrating",
+        "state": "published" if published and gated_available else "calibrating",
+        "gated_available": gated_available,
         "slo_id": "client_observed",
         "methodology_version": _optional_int(snapshot.get("methodology_version")),
         "windows": {
@@ -575,7 +579,7 @@ def client_observed_status_section(
         "canary": _status_canary(snapshot.get("canary")),
         # Tolerates a snapshot built by a worker that predates the calibration
         # view: the control plane deploys before the ClickHouse node does.
-        "all_traffic": _status_all_traffic(snapshot.get("all_traffic")),
+        "all_traffic": all_traffic,
         "generated_at": generated_at if isinstance(generated_at, str) else None,
     }
 
--- a/src/trusted_router/client_reliability.py
+++ b/src/trusted_router/client_reliability.py
@@ -14,6 +14,9 @@ ALL_TRAFFIC_NOTE = (
     "Calibration view; includes synthetic canary traffic; uncapped; ungated; "
     "not the published methodology."
 )
+#: Gated windows exclude synthetic traffic, so one 24h request is enough to
+#: distinguish real client use from an empty calibration-only snapshot.
+MIN_REAL_REQUESTS_24H = 1
 
 HOSTS = (
     "apex",
@@ -554,6 +557,10 @@ def client_observed_status_section(
     published = snapshot.get("published") is True
     raw_windows = snapshot.get("windows")
     windows = raw_windows if isinstance(raw_windows, Mapping) else {}
+    raw_24h = windows.get("24h")
+    window_24h = raw_24h if isinstance(raw_24h, Mapping) else {}
+    if _int(window_24h, "requests") < MIN_REAL_REQUESTS_24H:
+        return {"available": False, "reason": "insufficient_real_data"}
     generated_at = snapshot.get("generated_at")
     return {
         "available": True,
--- a/src/trusted_router/config.py
+++ b/src/trusted_router/config.py
@@ -394,6 +394,10 @@ class Settings(BaseSettings):
     # Durable tenant-activity and synthetic-status stream. Kept independent
     # from provider analytics so its privacy and cutover can be controlled.
     operational_analytics_outbox_enabled: bool = False
+    # Client-observed figures include upstream provider-caused failures and can
+    # understate a healthy gateway. Keep them off public surfaces until
+    # per-provider attribution makes the number fair to publish.
+    public_client_observed_enabled: bool = False
     # Client-observed reliability beacons remain off until the ClickHouse node
     # runbook has been completed. Every setting is available as TR_CLIENT_EVENTS_*.
     client_events_enabled: bool = False
--- a/src/trusted_router/routes/public.py
+++ b/src/trusted_router/routes/public.py
@@ -2217,9 +2217,7 @@ def _merge_client_observed_status(
     *,
     settings: Settings,
 ) -> dict[str, Any]:
-    result = _apply_public_client_observed_policy(payload, settings=settings)
-    if not settings.public_client_observed_enabled:
-        return result
+    result = dict(payload)
     snapshot = None
     if settings.environment != "test":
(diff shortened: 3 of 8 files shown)


jperla:
Refined after the owner asked why the Python-SDK synthetic numbers were not showing. They now exist: the `fleet_all` rollups landed at 13:30Z and the live snapshot reads `all_traffic 24h = 3,617 requests` while gated stays 0 — and the first version of this PR would have hidden exactly that.

Three states now, not two:

| gated (real) | all_traffic (incl. synthetic) | page |
|---|---|---|
| has requests | — | full section, unchanged |
| **empty** | **has requests** | **labelled calibration row only** — no `insufficient data` cards |
| empty | empty | nothing rendered, section element included |
| `reason == stale` | — | stale warning, unchanged |

`gated_available` names the state in the projection so the template branches on it and ops can distinguish calibration-only from nothing-at-all; `state` stays `calibrating` unless published **and** gated data exists.

Reviewer verification against the **live production snapshot**, rendering the real template fragment:

```
Client-observed availability
Calibration only; no customer traffic has been measured yet.   [calibrating]
24h all traffic — calibration view — includes synthetic canary traffic; not the published number
Availability 83.6922%   Requests 3617   p50 total 1600 ms   p95 total 12800 ms
```
`contains 'insufficient data' = False`, `contains 'calibration view' = True`.

Projection states on live data: today → `available=True, state=calibrating, gated_available=False`; both-empty variant → `available=False, reason=insufficient_real_data`; injected real traffic → `gated_available=True`.

Gates: **57 passed**, ruff and mypy clean. Reviewer mutations: render gated cards in calibration-only mode → `test_status_html_shows_only_calibration_without_gated_traffic` failed; never hide when both are empty → `test_status_html_hides_client_section_without_real_traffic` and the projection table test both failed; restored → 57 passed.

jperla:
Scope changed on the owner's decision: client-observed availability is **not published at all** — it stays internal. His reason: bad providers hurt the numbers. Production bears that out — the figure reads **83.69%** today while the gateway itself is healthy, because it counts upstream provider failures (and this morning's `responses_pong` 501s). Publishing it would state something untrue about the service.

`TR_PUBLIC_CLIENT_OBSERVED_ENABLED` (default **false**) removes the key from **every** public surface, not just the page — hiding only the HTML would leave the numbers one `curl` away:

- `/status` HTML — section absent
- `/status.json` — `client_observed` key absent, not present-and-empty
- the peer fan-out `/fleet` consumes — and `/fleet` is now tested against peers that omit the key

Cache keys include the flag, so a payload built while it was on cannot be served after it goes off. Untouched: the authenticated console, internal alerting (`client_watch.py` keeps detecting invisible outages), and the public methodology page — whose wording no longer claims the numbers are published. The earlier three-way rendering stays for the enabled path, so this is a disclosure decision, not a deletion of the work.

Reviewer evidence: **76 passed** across the client-reliability, public-status and fleet suites; ruff and mypy clean. An end-to-end check against the real app with default settings asserts no `client-observed` in the `/status` HTML and no `client_observed` in `/status.json` or `/fleet.json` — **passed**. Reviewer mutation: default the flag to `True` → `test_public_client_observed_flag_defaults_off_and_reads_env` and `test_public_status_surfaces_omit_client_observed_by_default` **both failed**; restored → 10 passed.

jperla:
Adversarial review **BLOCKED** the previous head on a finding I had missed, and it was right: the flag-aware cache keys only partition the **process-local** cache. Cloudflare fronts the status host (`scripts/deploy/cloudflare_lb.sh`) and keys by URL, so a body produced while client-observed was enabled remained servable to anonymous callers for `s-maxage=60` plus `stale-while-revalidate=600` after the flag went off — about eleven minutes of precisely the disclosure this PR exists to prevent.

Fixed on head `37adfc76`: with the flag **on**, `/status`, the status-host `/` and `/status.json` answer `no-store`, so no shared cache can hold one. With the flag **off** (default, normal state) the existing headers are unchanged and now asserted verbatim, so the status page keeps its edge caching and a future change to those headers is caught. A single helper picks the header so the three routes cannot drift.

Everything else in that review came back clean and is worth recording: direct origin paths remove the key rather than emptying it, snapshot cache hits and **both** stale-fallback paths reapply the policy, `/status/history` never touches client data, and — the subtle one — `/fleet` will not republish an **older peer's** `client_observed` because `_peer_summary` projects only banner, component, monitor and timestamp fields. No client-derived component, SLO, headline, freshness field, JSON-LD, OG image, feed or header was found. Console and internal alerting unaffected.

Evidence: **79 passed** across the three suites; ruff and mypy clean. Author mutation (emit the public caching header even when enabled) failed all three parametrised surfaces; restored → passed. Reviewer check against the real app: flag off keeps the public caching header and exposes no `client_observed`.

**Operational note for whoever deploys this:** bodies cached at Cloudflare *before* this change still contain the numbers. The code stops future leaks but cannot evict what is already at the edge — a one-time purge of `/status`, `/status.json` and the status host `/` is needed after rollout.

jperla:
Second block, and this one was a **consequence** rather than a code bug: removing `client_observed` from `/status.json` would have broken `.github/workflows/check-client-telemetry-freshness.yml`. That job runs on a 06:45/18:45 cron and reads the public `/status.json` **with no credentials on purpose** — its header calls it "the check that survives the watch itself being down", i.e. the outside-in detector for a dark telemetry pipeline. It would have reported `/status.json has no client_observed section` and filed an issue twice daily. A disclosure change would have silently disabled an outage detector.

The resolution: what that check consumes is **not statistics**. `evaluate()` reads `available`, `generated_at`, and `canary.{last_seen_age_seconds,last_24h_count}` — a pipeline heartbeat and the liveness of *our own* synthetic batches. Your objection is to reliability figures that count upstream provider failures.

So the public payload now carries **liveness only**, as an explicit allowlist (`available`, `reason`, `generated_at`, `canary`), and every statistic stays private: availability percentages, request/success/fault counts, tenant counts, latency percentiles, per-host and per-SDK breakdowns. Naming what may be published means future snapshot fields are private by default. The page still renders nothing — the template gates on the flag, not on the key.

Reviewer verification on this head: **80 passed** across the three suites, ruff and mypy clean. Two independent checks against the real app — the public `client_observed` contains only the four allowlisted keys with no statistic string anywhere in it and the HTML has no client-observed section; and the **real** `clickhouse.check_client_telemetry_freshness.evaluate` accepts the flag-off payload built by the app (no "no client_observed section" problem). Author mutation (include `windows` in the liveness projection) failed `test_public_status_surfaces_publish_only_client_liveness_by_default`; restored → passed.

PR #742, opened 2026-08-22
Give the T1 public website its own service, secrets and edge policy

jperla:
Gives the **T1 public website** its own process, secrets and edge policy, so an incident in T2–T4 cannot take the public site down.

Josh's availability tiering, which is the spec here:

| Tier | | Served by, after this |
|---|---|---|
| **T1** | public web pages — **must stay up** | **new `trusted-router-public`** |
| T2 | group buy — ok to go down | legacy (unchanged) |
| T3 | signup | legacy (unchanged) |
| T4 | internal metrics; logged-in user parameters | legacy (unchanged) |
| T5 | core routing — must be super stable | separate infrastructure, untouched |

Dependency rule: a lower tier may depend on a higher one, never the reverse. The console loading `/static` from T1 is allowed; T1 depends on nothing below it.

## Why this direction

The `public` surface already boots from today's image with a small secret set and **refuses at boot** to hold the gateway token, Stripe keys, Sentry (before this PR) or the observer token. Isolation is enforced fail-closed, so no new isolation code was needed. The reverse split is not possible: a `control` process refuses to boot with the internal-gateway/synthetic secrets and mounts no `/internal/gateway/*` or `/support/inquiry`, which nothing else on main serves — so the legacy service must stay `combined`.

## What it does

- `scripts/deploy/public_surface.sh` — deploys the service in four regions from the same image, in two stages: `companion` (ingress all, `TR_RATE_LIMIT_CLIENT_IP_MODE=untrusted`, no user traffic) and `routed` (ingress internal-and-LB, `edge_header`). The mode must not be `edge_header` while the service is reachable on run.app, since the limiter trusts that header with no source check.
- `scripts/deploy/public_surface_edge.sh` — per-region NEGs, a new backend with CDN and the client-IP header, **attaches** a pre-existing Cloud Armor policy, and separate `cutover` / `rollback` subcommands. Rollback restores a map captured **before** the import, verified by digest, and refuses to re-render.
- CI deploys and smokes the companion only. **The cutover is never run by CI** — it is an owner action.
- Two app changes only: Sentry becomes public-owned (T1 error reporting), and bare `/bedrock-group-buy` joins its `/*` siblings on control so T2 is not split across tiers.

## Verification

- Every path the URL map sends to public must be mounted by the public app — **currently zero violations** (public mounts 299 routes, combined 642). This is what stops a future route silently 404-ing after cutover.
- Reviewer-run: 338 tests pass; author-run: full suite 5935 passed, mypy, ruff, actionlint, `bash -n` clean.
- Verified against live prod by the reviewer: the new legacy deploy gate expects `/console`=302 and `/auth/session`=401, which **matches production today**, so it will not false-trigger.
- Verified the Sentry ownership change did not widen anything else: public still refuses the gateway token, Stripe key, Stripe webhook secret, BYOK key and observer token.
- Confirmed no `security-policies create|update` is ever executed (`tr-deploy@` lacks that permission); those commands are printed as owner instructions, and their host regex uses non-capturing groups because Cloud Armor's RE2 rejects capture groups.

## Blocking before deploy

The deploy binds `TR_ATTRIBUTION_COOKIE_SECRET=trustedrouter-attribution-cookie-secret:latest`, **which does not exist in Secret Manager** — the deploy would fail. The legacy service binds no attribution secret: it derives the cookie key from the gateway token, which `public` refuses by design. Decision taken: share only the **derived** key (`HMAC-SHA256(gateway_token, "trustedrouter-attribution-cookie-v1")`) — it reproduces today's key exactly so no cookie breaks, and is one-way so it grants no gateway access. Verified: the derived value matches `_cookie_signing_key` byte-for-byte, and the encryption key chains off the signing key, so matching it gives full cookie compatibility. A follow-up commit adds `TR_ATTRIBUTION_COOKIE_KEY` and switches this binding.

Also pending: `tr-public@` exists but has zero role grants, and the T1 Cloud Armor policy must be created by the owner. Both are in the prerequisites runbook.

## Known limits

- The legacy service keeps `ingress=all` and its run.app URL, so this isolation is routing-level until that is closed separately.
- 14 other `/x/*`-without-bare-`/x` patterns have the same fail-open shape as group buy did (none are mounted routes today, so they 404 either way). A follow-up makes the matcher handle it systemically.

Written by codex; reviewed by Claude / Fable.


jperla:
**Both blockers cleared** (commit `60b4e312`, codex-authored, Fable-reviewed).

**Attribution key.** The deploy now binds `trustedrouter-attribution-cookie-key:latest`. `TR_ATTRIBUTION_COOKIE_KEY` holds base64 of `HMAC-SHA256(gateway_token, "trustedrouter-attribution-cookie-v1")` and is used directly as the key. Reviewer-verified independently: signing keys match the legacy process **and so do encryption keys** (the encryption key chains off the signing key, so matching the signing key alone would not have been sufficient to assume). Fail-closed guards confirmed by probe: non-32-byte value rejected, key+secret both set rejected, and key equal to the gateway token rejected in production.

**Tier fail-open in the matcher.** The bare `/bedrock-group-buy` hole was one of **fifteen** — a `/x/*` pattern did not match bare `/x`, which fell through to the *public* default, so an unlisted path silently landed in the must-stay-up tier. Fixed once in the matcher rather than by 14 bare twins. Reviewer-measured across **all 521 concrete routes**, comparing this branch against main: **exactly three change tier**, all group-buy variants moving public → control, which is the intended T2 outcome. **Zero** bare-vs-glob holes remain, and the "every path the map sends to public is mounted by public" guard still reports zero violations.

Reviewer-run tests: 397 passed. Author-run full suite: 5956 passed, mypy and ruff clean.

Still owner-run before any deploy, per the prerequisites runbook: create `trustedrouter-attribution-cookie-key` (one pipeline, gateway token never leaves the pipe), grant `tr-public@` its roles (it exists with zero grants today), and create the T1 Cloud Armor policy (`tr-deploy@` lacks `securityPolicies.create` by design). PR stays draft until those land.

jperla:
**All owner prerequisites are done and verified in production** (run by Fable at Josh's request; none touched serving traffic).

- `trustedrouter-attribution-cookie-key` created — stores 44 base64 chars decoding to exactly 32 bytes. **Proven against the live secrets, with no value printed: the stored key reproduces the legacy signing key exactly (`base64decode(key) == HMAC-SHA256(gateway_token, "trustedrouter-attribution-cookie-v1")` → True), and is not the token itself → True.** So attribution cookies survive the split with zero breakage.
- `tr-public@` granted `spanner.databaseReader`, `bigtable.reader`, `serviceusage.serviceUsageConsumer`, and `secretmanager.secretAccessor` on exactly the three secrets it needs — nothing else.
- `tr-deploy@` granted `iam.serviceAccountUser` on `tr-public@`.
- Cloud Armor `trusted-router-public-edge` created with adaptive protection on, rule 900 host-deny (non-capturing groups) and rule 1200 per-source ceiling 2400/min → 429, both enforcing. Created **unattached** — nothing routes through it until cutover.
- Production unaffected throughout: `/v1/health` and `/` both 200 in ~120–150 ms.

Correction to the earlier runbook: the role is `roles/secretmanager.secretAccessor`, not `roles/secretAccessor` (which does not exist). Fixed before running.

CI is green 8/8 on `60b4e312`. Merging this runs the **companion** deploy only — a new service taking zero user traffic; the URL-map cutover stays a separate owner-run command.

PR #777, opened 2026-08-23
Install markdown parser in production image

jperla:
Production revision trusted-router-01200-dqv failed before serving traffic with ModuleNotFoundError: No module named 'bs4'. Public markdown negotiation imports BeautifulSoup at runtime, but beautifulsoup4 was only in the dev dependency group while the Cloud Run image installs with --no-dev.\n\nThis moves beautifulsoup4 into project runtime dependencies, refreshes uv.lock, and adds a regression guard.\n\nVerified:\n- 32 focused markdown/discoverability/runtime dependency tests pass\n- frozen --no-dev export contains beautifulsoup4==4.14.3\n- production-only environment imports trusted_router.markdown_negotiation\n\nNo traffic shifted to the failed revision; production stayed on trusted-router-01199-649.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,8 @@ requires-python = ">=3.11"
 license = "BUSL-1.1"
 dependencies = [
   "axiom-py>=0.9.0",
+  # Public markdown negotiation parses rendered HTML at request time.
+  "beautifulsoup4>=4.12.0",
   "boto3>=1.35.0",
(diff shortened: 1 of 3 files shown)


jperla:
Superseded by #778, which moved beautifulsoup4 into runtime dependencies and added an explicit production-image import smoke check. Closing this duplicate to keep the fix path singular.

PR #779, opened 2026-08-23
Move beautifulsoup4 to runtime deps; smoke the no-dev import in CI

jperla:
Main is currently undeployable: every deploy since #764 fails at the
us-central1 traffic ramp because revision containers crash on boot with
`ModuleNotFoundError: No module named 'bs4'`
([markdown_negotiation.py:46](src/trusted_router/markdown_negotiation.py#L46), imported at startup via
middleware → dashboard → main). Production keeps serving the old
revision — the staged ramp refused to shift traffic to a revision that
never became ready, which is it working as designed.

Why CI never saw it: `beautifulsoup4` sat in the **dev** dependency
group with a comment promising it was test-only, every CI job syncs the
dev group, and the production image installs `--no-dev`
([Dockerfile:14-17](Dockerfile#L14)). All tests green; container cannot boot.

Two changes:

1. **beautifulsoup4 moves to `[project] dependencies`** (with the
   incident written into the comment where the stale promise used to
   be), `uv.lock` re-resolved — a 2-line lock diff.
2. **`no-dev-import-smoke` CI job**: `uv sync --frozen --no-dev` then
   `python -c "import trusted_router.main"` — boots the import graph
   exactly the way uvicorn does in the image, so the entire class of
   "runtime module imports a dev-only package" fails in CI instead of
   at the production ramp.

Verified locally: the import fails on main under `--no-dev` and
succeeds with this change; full dev sync restored afterwards.

This also unblocks #771's fix (T1 outbox flag), which merged but cannot
reach production while deploys fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,27 @@ jobs:
   # tests/test_check_format_ordering.py, in
   # test_an_unchanged_write_side_is_silent_and_reads_no_network, which drives
   # the real main() with every network seam replaced by a function that raises.
+  no-dev-import-smoke:
+    # The production image installs with --no-dev (Dockerfile), while every
+    # other CI job syncs the dev group. A runtime module importing a dev-only
+    # package therefore passes ALL tests and then crashes the container at
+    # startup, failing every deploy at the traffic ramp — which is exactly
+    # what happened when markdown_negotiation imported bs4 while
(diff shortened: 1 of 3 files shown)


jperla:
Superseded by #778, which landed the identical fix concurrently — beautifulsoup4 into runtime deps plus a --no-dev import smoke in CI. Nothing here survives that isn't already on main.

PR #780, opened 2026-08-23
Fix production Markdown runtime dependency

jperla:
## Summary
- move Beautiful Soup into production dependencies because Markdown negotiation imports it at app startup
- add a CI job that installs only production dependencies and imports the runtime Markdown module
- refresh the frozen lockfile

## Incident
The zero-traffic Cloud Run canary `trusted-router-01200-dqv` exited before listening on port 8080 with `ModuleNotFoundError: No module named 'bs4'`. Existing production traffic stayed on `trusted-router-01199-649`.

## Verification
- production-only `uv sync --frozen --no-dev`
- production imports: `bs4`, `trusted_router.markdown_negotiation`
- `uv run ruff check .`
- `uv run pytest -q tests/test_markdown_negotiation.py` (20 passed)
- `uv lock --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,21 @@ on:
 # conclusion for the SHA, and a run only concludes success when every job
 # succeeds, so the gate semantics are unchanged by the split.
 jobs:
+  runtime-import:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


jperla:
Closing as a duplicate of #778, which merged the same production dependency promotion and a stronger full-app production-only import gate. This PR's CI confirms both production import jobs pass against current main.

PR #782, opened 2026-08-23
Repair stale failed Cloud Run traffic before deploy

jperla:
## Summary
- detect failed Cloud Run revisions that remain referenced by desired traffic or a stale probe tag
- restore desired traffic to the sole effective 100% healthy revision before pruning
- verify the live state after mutation, including when gcloud reports a misleading nonzero result
- fail closed when effective traffic is split or ambiguous

## Incident
Revision `trusted-router-01200-dqv` failed app startup. Cloud Run correctly kept effective traffic at 100% on `trusted-router-01199-649`, but desired traffic retained a 90/10 split and a stale `staged-probe` reference. The next deploy created healthy revision `trusted-router-01201-5nk` but still returned the old revision's failure.

## Verification
- `bash -n scripts/deploy/rollout.sh`
- `uv run ruff check .`
- 37 focused deploy/helper tests passed
- malformed state and active split tests prove the repair remains fail-closed

### DIFF
--- a/scripts/deploy/resolve_failed_revision_repair.py
+++ b/scripts/deploy/resolve_failed_revision_repair.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+"""Plan repair of failed Cloud Run revisions left in service traffic state."""
+
+from __future__ import annotations
+
+import json
+import sys
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
+from typing import Any
+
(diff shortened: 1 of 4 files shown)


jperla:
Closing as superseded by #781, which fixes the same stale desired-traffic failure with a smaller normalizer and explicit staged-update rollback. The current main deployment is exercising #781 now; keeping both implementations would duplicate deployment state logic.

PR #790, opened 2026-08-24
Hold Sakana routes under global region policy

jperla:
## Summary
- keep Sakana Namazu and Fugu visible in the provider manifest while preventing either from entering authorization
- add a code-level Namazu hold that survives stale or incorrectly relisted manifests
- record the geographic-policy hold in the generated manifest

## Root cause
Sakana documents that its API is unavailable in the EEA, UK, and Switzerland and may enforce this with IP controls. Production smoke proved Namazu succeeds from us-central1, us-east4, and southamerica-east1, but Sakana returns an HTML 403 from both europe-west4 enclave addresses. Because api.trustedrouter.com currently publishes all regional enclave IPs in one DNS answer, a region-only filter would still make canonical calls fail nondeterministically.

## Verification
- `uv run ruff check ...`
- 215 focused catalog/provider tests passed
- `uv run mypy`
- direct Sakana call and three non-EU regional TrustedRouter PONGs passed before hold
- Europe West reproduced provider 403 before hold

Claude CLI could not perform the follow-up review because the local CLI is signed out. The original Sakana integration received an Opus `SAFE TO MERGE` review; this hotfix was independently reviewed after the focused suite.

### DIFF
--- a/scripts/pricing/providers/sakana.py
+++ b/scripts/pricing/providers/sakana.py
@@ -15,6 +15,8 @@ from scripts.pricing.providers._direct_openai import DirectOpenAIProvider, Direc
 from trusted_router.provider_contracts import (
     SAKANA_FUGU_MODEL_ID,
     SAKANA_FUGU_ROUTE_HOLD_REASON,
+    SAKANA_NAMAZU_MODEL_ID,
+    SAKANA_NAMAZU_ROUTE_HOLD_REASON,
 )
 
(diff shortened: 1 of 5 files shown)


jperla:
Full CI later exposed two stale provider social-card assertions after the route hold changed Sakana's active counts from 1 to 0. Follow-up #791 regenerates the committed card and manifest; focused branding and Sakana tests pass (69/69).

PR #799, opened 2026-08-24
ci: 8-core test job (draft until the runner proves it schedules)

jperla:
Probe PR: the test job is ~23m on 4 vCPUs and is the long pole of gate-on-ci, which every deploy waits on. This moves it to `ubuntu-latest-8-cores` with `-n 8`.

**Draft on purpose.** If the org has no 8-core runner group configured, this label queues forever — merged, that would stall every deploy's gate fleet-wide. The draft run answers two questions safely: does the label schedule at all, and what does the wall-clock drop to. Merges only after both answers are in (and after the bake-ladder PR, to keep the merge queue single-file).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,12 @@ jobs:
         run: uv run --no-sync python -c "import trusted_router.main"
 
   test:
-    runs-on: ubuntu-latest
+    # Larger runner: this job is ~23 minutes on 4 vCPUs and sits on the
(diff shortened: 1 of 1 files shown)


jperla:
Probe answered: `ubuntu-latest-8-cores` never scheduled (12 min queued) — no larger-runner group is configured for this org. This is exactly what the draft guard existed to catch: merged, this label would have stalled gate-on-ci on every deploy. The test-job speedup proceeds via pytest sharding across standard runners with a coverage-combine job instead.

PR #810, opened 2026-08-24
Cloud bake ladder: canary one cloud fast, promote after a real bake

jperla:
The deployment model Joseph specified, made enforceable rather than
conventional:

| You want | You run | The gate demands |
|---|---|---|
| Test fresh code in prod NOW | `TR_CLOUD_DEPLOY_MODE=canary bash scripts/deploy/azure_control_plane.sh` | one OTHER **gated** cloud serving a ≥24h-baked commit (the lifeboat), fleet up |
| Roll it to another cloud | default mode | candidate **merged** ≥24h ago AND its lineage carrying traffic on a known cloud now |
| Emergency everywhere | `TR_CLOUD_BAKE_OVERRIDE="<reason>"` | prints every real check result, logs the reason, proceeds |

Fresh code on all clouds at once is structurally impossible; clouds on
different commits for days is the designed state.

## What makes the guarantee real (adversarial review, 9 findings, all fixed + regression-tested)

- **Discovery reads traffic, not configuration.** GCP: the active
  traffic-weighted revision via `resolve_active_revision.py` — the
  service *template* lies after every rollback, since rollback paths
  revert traffic only. AWS: only trusted when `Service.Status=RUNNING`
  and the latest operation `SUCCEEDED`. Azure: the traffic-carrying
  revision, not the app template. A rolled-back deploy cannot testify
  for its own commit.
- **"Baked" means merged-age, not commit-age.** Containment in
  `origin/main` history that is itself ≥24h old — `GIT_COMMITTER_DATE`
  backdating buys nothing.
- **GCP cannot be the lifeboat** — its auto-deploy never runs this
  gate, so its baked status can vanish on the next routine merge.
- **The artifact is gated, not just HEAD**: dirty tree and
  tag≠HEAD both refuse (a mislabeled tag would launder unbaked code
  into future lineage evidence).
- **Ref-shadowing killed**: a branch named like a short sha cannot
  redirect `rev-parse` (empirically reproduced before the fix).
- **Abort is proven, not assumed**: harness runs without the override
  show rc≠0, zero build/deploy CLI calls after refusal, and the mutex
  released through the EXIT trap.

## GCP deploys get faster without touching #695

Secondaries **warm in parallel** (no-traffic revisions; no shared-state
mutation — LB reconciliation stays primary-only) and **ramp serially**,
preserving the 2026-08-20 billing-contention containment verbatim.
Secondary phase ~30m → ~12m; whole deploy ~45m → ~25m. Every region
keeps its watchdog, billing-5xx gate (window now stamped per-ramp), and
independent rollback.

Also: the fleet mutex holder record gains a `cloud` field; both
operator scripts take the mutex with the workflow's trap discipline;
`azure_canary_app.sh` refuses production resource names; AWS `TAG`
defaults to the short sha (was mutable `eu`) with digest-pinned serving
unchanged.

## Verification

ruff · mypy · **6287 passed** full suite · 278 focused
(gate/mutex/workflow/harness/synthetic) · bash -n + ShellCheck clean on
all five changed scripts (remaining notes are pre-existing SC1091
info). Runbook documents both modes, the lifeboat table, and when the
override is legitimate.

Post-merge verification planned: stopwatch a real deploy for the
~25m target, then provoke a real bake-gate refusal on the AWS script
(the gate exits before any mutation, so the refusal is free).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -386,9 +386,10 @@ jobs:
       # Staged regional deploy. Deploy us-central1 first, gate on a
       # 3-min single-region synthetic watch; only proceed to the
       # secondary warm regions if us-central1 stays up. Once the image has
-      # passed that primary-region gate, every secondary rolls sequentially.
-      # All regions share Spanner, so overlapping revision warmups and traffic
-      # ramps can amplify billing transaction contention.
+      # passed the primary's full ramp and 3-minute canary, secondary revisions
+      # warm in parallel without traffic, then ramp serially. Incident #695
+      # (billing 5xx, 2026-08-20) showed that overlapping revision warmups and
+      # traffic ramps can amplify billing transaction contention.
       #
       # Hotfix mode (workflow_dispatch input `hotfix: true`): the
       # watchdogs skip baseline + use --skip-baseline so a sick prod
@@ -536,7 +537,7 @@ jobs:
             exit 1
           fi
 
-      - name: Roll secondary warm regions sequentially
+      - name: Warm secondaries in parallel, ramp serially
         env:
           IMAGE_TAG: ${{ needs.build-image.outputs.sha }}
           PREV_EU: ${{ steps.prev.outputs.europe_west4_revision }}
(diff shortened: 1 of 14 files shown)


jperla:
GitHub did not spawn the `pull_request` CI run for this branch (same known glitch as #602/#608). CI was dispatched explicitly and concluded **success, 8/8 jobs**: https://github.com/Lore-Hex/quill-router/actions/runs/32790032567 — merging on that run's green. `gate-on-ci` will independently verify the merge commit before any deploy mutation.

jperla:
Post-rebase CI green: https://github.com/Lore-Hex/quill-router/actions/runs/32792540299 (rebase reconciled #768's RELEASE_COMMIT staleness detector with the immutable sha tags: RELEASE_COMMIT now aliases the tag, one source of truth, and the silent 'unknown' fallback became a hard refusal).

PR #815, opened 2026-08-25
Primary-live deploys: warm everything early, converge behind the same lock

jperla:
Joseph's levers 1+2: free overlaps and a job split. No watchdog window,
ramp step, or canary second changes.

## Shape

**`deploy` (primary-live, ~9.5-11m):** acquire mutex → warm ALL FOUR
regions' no-traffic revisions in parallel (the primary's deploy was
always just a warmup; LB reconciliation stays primary-only) → ramp
us-central1 10/50/100 unchanged → 3-min canary unchanged → success.
This is the moment new code serves.

**`rollout-secondaries` (needs: deploy):** imports the mutex fence from
job outputs, ramps the three already-warm secondaries serially exactly
per #695, runs the reconciler deploy in the background alongside,
smokes, releases the lock in `always()`.

**Release-ownership matrix** (commented in the workflow, pinned by
tests): primary fails → primary's conditional release; primary succeeds
→ secondaries' `always()` release; secondaries never scheduled → the
90-min TTL recovers (freeze, never overlap). Timeouts 25+55 < TTL, the
arithmetic stated inline. The lock spans both jobs, so another cloud's
canary still fail-closes mid-convergence.

## The review catch

Codex's draft piped the acquire: `acquire | tee -a $GITHUB_ENV
$GITHUB_OUTPUT`. GitHub's default run shell is `bash -e` **without
pipefail** — a BLOCKED acquire would exit 0 through tee and the deploy
would proceed without holding the lock. Its own structural test had
pinned the broken construct verbatim. Fixed to command substitution
(exit code propagates; the fence reaches both files only on success),
with two tests standing guard: the pin asserts the safe form, and a
second fails on ANY pipe after `deploy_mutex.sh acquire`.

## Measured context

First live ladder run (32794630988): primary lane 12m19 (deploy 1m35 +
ramp 7m13 + canary 3m31), secondaries 27m58. The ramp overrun is
overhead around the windows (~31s watchdog startup per step, one 3m07
Cloud Run traffic-update stall), not the windows themselves — follow-up
material, not this PR.

## Verification

ruff · mypy · full suite **6414 passed** (one apparent failure was a
stale shared-venv artifact in my test harness, not the diff — 21/21
green with an honest venv) · focused workflow/mutex/bake suites 100
passed · YAML parses · runbook + multi-cloud-separation docs updated.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -345,11 +345,20 @@ jobs:
     needs: [build-image, migrate-schema, sync-runtime-secrets, confirm-current-main]
     if: ${{ needs.confirm-current-main.outputs.proceed == 'true' }}
     runs-on: ubuntu-latest
-    # Must stay UNDER the deployment mutex TTL (deploy_mutex.sh, 5400s): a job
-    # that outlives the lease invites a legal expired-lock takeover while this
-    # run is still shifting traffic — the exact overlap the mutex exists to
-    # prevent. GitHub's default 360-minute bound is four TTLs too generous.
-    timeout-minutes: 85
+    outputs:
+      TR_DEPLOY_MUTEX_OPERATION: ${{ steps.acquire_mutex.outputs.TR_DEPLOY_MUTEX_OPERATION }}
+      TR_DEPLOY_MUTEX_GENERATION: ${{ steps.acquire_mutex.outputs.TR_DEPLOY_MUTEX_GENERATION }}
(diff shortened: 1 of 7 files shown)


jperla:
CI green: https://github.com/Lore-Hex/quill-router/actions/runs/32803135525 (dispatched; the pull_request trigger flaked again). Merging — the deploy of this merge is the split's first live run and will be measured.

PR #823, opened 2026-08-25
Reconcile the usage baseline instead of giving up on it

jperla:
The first production run of the usage audit reported `usage: 0/643 (643 unauditable)` — it covered nothing. I told you that meant every baseline was gone and the feature needed a delta redesign. I queried production instead of inferring, and the second half of that was wrong.

## What production actually says

The JSON baselines really are gone — 643 credit rows, **zero** with any of `total_usage_microdollars`, `total_credits_microdollars`, `reserved_microdollars`. The cleanup ran fleet-wide.

But absent-in-JSON is not unrecoverable, because a baseline is usage booked *before* the ledger and therefore can never be negative. Two cases follow from that alone:

| | workspaces | |
|---|---|---|
| Ledger explains the counter **exactly** | **604** | baseline is zero — reconciled, not assumed |
| Counter exceeds the ledger | **39** | genuinely pre-ledger history, $9,442 total |
| Ledger exceeds the counter | **0** | — |

That last row is the reassuring one: **no workspace has booked spend its counter never received.** Every discrepancy points the other way, which is what a lost pre-flip baseline looks like. The $9,442 is also extremely concentrated — one workspace is $8,459 of it.

## What changes

The audit resolves a baseline three ways — a recorded row, then the JSON key, then **zero when the ledger accounts for the counter exactly**. That turns 94% of the fleet from uncheckable into guarded, without assuming anything.

It also gains a check it could not make before: **a counter below its ledger is a violation with or without a baseline**, since no baseline can be negative. Previously "no baseline" ended the check and money booked-but-not-counted would have been invisible.

For the 39, `scripts/record_usage_baselines.py` records the remainder the ledger cannot account for. It invents no number — that remainder *is* the pre-ledger usage — and it changes no counter and moves no money. Dry-run by default, `--apply` gated on the Spanner backend, refuses to overwrite an existing baseline, and stores the evidence it was derived from beside the value.

The row lives under its own entity kind rather than back in the credit body, because that body is exactly what `storage_gcp_credit_json_cleanup` empties. Putting it back there would arrange the same loss a second time.

## Verification

Mutation-tested, and it caught one of mine: removing the reconciled-zero rule and removing the below-ledger violation each turn a test red — but removing the **in-transaction** overwrite guard did not, because an early return masked it. That test was vacuous. Added the race case (a baseline recorded between propose and write) that actually exercises it; the mutation now bites.

76 tests pass across the billing/auditor/cleanup suites; ruff and mypy clean.

**No production write in this PR.** Recording the 39 baselines is an operator step through the new CLI, per the repo's rule that a new prod operation ships as reviewed tooling first. Preflight it with a bare `uv run python scripts/record_usage_baselines.py` — it prints the list and the total before anything is written.

### DIFF
--- a/scripts/record_usage_baselines.py
+++ b/scripts/record_usage_baselines.py
@@ -1,0 +1,90 @@
+"""Record the pre-ledger usage baseline for workspaces that still need one.
+
+Read-only unless ``--apply``. It records no number of its own: the baseline is
+the remainder the typed ledger cannot account for, which is by definition usage
+booked before the ledger began. Recording it changes no counter and moves no
+money -- it makes ``audit_typed_invariants`` able to check that workspace at
+all, where today it reports it as unauditable.
+
+Most workspaces need nothing. Where the ledger already explains the counter the
+baseline is zero and no row is written; only a counter that EXCEEDS its ledger
+has history to record.
+
+  uv run python scripts/record_usage_baselines.py            # preflight
+  uv run python scripts/record_usage_baselines.py --apply
+
+Refuses to overwrite an existing baseline. A baseline is a claim about history,
+so a second, different value for one workspace means one of them is wrong and a
+person should decide which.
+"""
+
+from __future__ import annotations
+
+import argparse
+import datetime as dt
+import os
+import sys
+from typing import Any
+
+from trusted_router.config import Settings
+from trusted_router.storage import create_store
+from trusted_router.storage_gcp_counter_reconcile import (
+    propose_usage_baselines,
+    record_usage_baseline,
+)
+
+
+def _usd(microdollars: int) -> str:
+    return f"${microdollars / 1_000_000:,.2f}"
+
+
+def main(argv: list[str] | None = None, *, store: Any | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--workspace", action="append", dest="workspaces")
+    parser.add_argument("--apply", action="store_true")
+    args = parser.parse_args(argv)
+
+    if args.apply and os.environ.get("TR_STORAGE_BACKEND") != "spanner-bigtable":
+        print("ERROR: --apply requires TR_STORAGE_BACKEND=spanner-bigtable", file=sys.stderr)
+        return 2
+
+    active_store = create_store(Settings()) if store is None else store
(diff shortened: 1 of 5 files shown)


jperla:
**Ran the preflight against production and it caught a bug of mine before it wrote anything.** Recorded 13 baselines, not the 42 the first version proposed.

## The bug

`typed_usage` sums `total_usage` over **every** shard of a workspace, but the ledger sum filtered `ws_shard=0`. So every non-zero-shard settle looked like usage the ledger could not account for — which reads as pre-ledger history. Production has **638,484 settled reservations off shard 0 across 30 workspaces**.

The preflight proposed 42 workspaces needing a baseline. The corrected query says **13**. Twenty-nine of those would have been fictitious history written into billing metadata. The fleet-wide ledger no longer filters shard, and the fake now *refuses* a query that reintroduces the filter.

The per-workspace form keeps its filter deliberately — its only caller refuses sharded workspaces, and the filter makes it fail safe if one slipped through: booked comes out low, the target falls below the counter, and the monotonic guard refuses rather than lowers.

## What was actually recorded

13 baselines, $8,361.51, dominated by one workspace at $8,330.20 (`45819281…`, 1.1M settled reservations — a genuinely heavy user whose pre-flip history is real). The rest: $23.64, $4.99, $1.30, $1.05, and eight under a cent.

Coverage went from **0/643 uncheckable** to **0/660 clean**.

## The second thing production taught

The below-ledger check is now **reported, not faulted**. A settle credits the balance and marks its reservation in *separate transactions*, so a continuously-loaded workspace is permanently a little behind. One monitoring workspace read −3780, −1537, −856, −173 microdollars on consecutive snapshots while another sat at exactly **−9 in every single one**.

A second confirming snapshot separates most of them — a real gap holds its value while an in-flight one moves — but not reliably, since two snapshots can coincide. Failing the nightly audit on a fluctuating fraction of a cent is the cry-wolf shape this repo has already paid for once, so the count is surfaced beside the summary and `behind_ledger_clean` is reported separately from `clean`.

Verified across three consecutive production runs: **exit 0, CLEAN, usage 0/660**, with the behind-ledger count visibly moving between 1 and 2 — which is the honest picture.

`84cbb802…` at a stable −9 microdollars across every read is the one real under-count. It is $0.000009 and I have not chased it further; it is now visible, which it was not before.

PR #827, opened 2026-08-25
Deploy overhead package: overlap the waits, pre-warm the capacity, measure every run

jperla:
Approved package (Joseph): cut the measured 41m48 full-convergence wall
by attacking overhead only — every watchdog window, ramp step, and
canary second is unchanged.

## The five fixes, from the profiled logs

1. **Watchdog grace overlaps the shift.** The ~30s baseline startup that
   sat serialized after every one of 9+ shifts now runs DURING the
   shift; a gate file releases the polling loop only after
   `update-traffic` returns, so the full configured window runs
   post-shift by construction.
2. **Capacity pre-warm kills the 100% stall** (3m04 observed). The
   candidate revision carries a revision-level `--min-instances` equal
   to the service `--min`; the probe tag keeps it active at 0% traffic.
   gcloud semantics verified against the real CLI: service min is
   divided among traffic-receiving revisions (exactly why the stall
   existed), effective capacity is max(service, revision) so
   convergence is cost-neutral, and rollback deactivates it — no
   restore step to forget.
3. **Probe-tag setup hoisted into warm** (~45s × 4 regions).
4. **Primary ramps on its own warm** (1m35, not the 3m28 slowest
   secondary); secondary warms continue across step boundaries via
   atomic status files + PIDs, collected by an always() end-of-job wait
   that gates success and feeds the revision outputs.
5. **Every deploy self-reports** `full_convergence_seconds` and
   `primary_live_seconds` — step summary + a structured Cloud Logging
   line (`tr-deploy-metrics`) — so the number that matters is monitored
   always.

## Adversarial round: six findings, all fixed here

- **Stranded pre-warm capacity (MEDIUM):** an aborted workflow left
  tagged candidates holding revision minimums (up to 12 paid instances)
  until some future deploy. rollout.sh gains the INT/TERM→EXIT funnel,
  and both jobs' failure paths sweep `staged-probe` tags from
  never-ramped regions (idempotent).
- Collector TOCTOU (spurious red when a warm finishes between the
  status check and `kill -0`): the status FILE is the truth, re-checked
  after a settle beat.
- Unbounded status poll (PID reuse → generic job timeout): 15-minute
  per-region deadline with a named error.
- Metrics step can no longer redden a converged deploy
  (`continue-on-error`).
- The baked-revision-minimum caveat is documented at the source:
  lowering a region's min now requires a redeploy to take effect, and
  the max-not-sum platform assumption gets one manual billing check
  after the first prod run.
- **Verdict semantics are now executed, not asserted:** the harness
  watchdog stub honors the gate with real wall-clock waits, and two
  behavioral tests prove an unhealthy verdict rolls traffic back and
  the gate releases only after the shift. (Also incidentally exercised:
  the wrong-surface probe rollback, via a misconfigured first draft of
  the test.)

## Expected

~41m48 → **~24-25m** full convergence (primary-live ~9m inside it),
measured by fix 5 on the next real deploy. The pending 3m→1m decision
on per-secondary post-100% watches is worth a further ~6m if taken.

## Verification

ruff · mypy · **6536 passed** full suite · 24/24 watchdog harness
(behavioral) · bash -n + shellcheck clean (pre-existing info notes
only) · YAML parses · gcloud flag semantics verified against the
installed CLI, not memory.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -348,14 +348,16 @@ jobs:
     outputs:
       TR_DEPLOY_MUTEX_OPERATION: ${{ steps.acquire_mutex.outputs.TR_DEPLOY_MUTEX_OPERATION }}
       TR_DEPLOY_MUTEX_GENERATION: ${{ steps.acquire_mutex.outputs.TR_DEPLOY_MUTEX_GENERATION }}
+      TR_DEPLOY_MUTEX_CREATED_AT: ${{ steps.acquire_mutex.outputs.TR_DEPLOY_MUTEX_CREATED_AT }}
       sha: ${{ needs.build-image.outputs.sha }}
       image: ${{ needs.build-image.outputs.image }}
+      primary_live_seconds: ${{ steps.primary_live.outputs.primary_live_seconds }}
       previous_europe_west4_revision: ${{ steps.prev.outputs.europe_west4_revision }}
       previous_us_east4_revision: ${{ steps.prev.outputs.us_east4_revision }}
       previous_southamerica_east1_revision: ${{ steps.prev.outputs.southamerica_east1_revision }}
-      europe_west4_revision: ${{ steps.warm_all.outputs.europe_west4_revision }}
-      us_east4_revision: ${{ steps.warm_all.outputs.us_east4_revision }}
-      southamerica_east1_revision: ${{ steps.warm_all.outputs.southamerica_east1_revision }}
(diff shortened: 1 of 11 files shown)


jperla:
CI green after the hermetic-harness fix: https://github.com/Lore-Hex/quill-router/actions/runs/32862116253 (first run failed only on CI: the harness inherited rollout.sh's operator-laptop TRUST_FILE default — now pinned to a harness-owned fixture, and the trust parse tolerates non-object JSON).

PR #833, opened 2026-08-25
confidential-cowork: tell people how to actually open the download

jperla:
## Why

The macOS build the page links to is ad-hoc signed and not notarized. `spctl -a -t exec` on the published app returns `rejected`, so macOS refuses the first launch with *"Apple could not verify 'Confidential Cowork' is free of malware that may harm your Mac or compromise your privacy."*

The page gave the user nothing to work with: no install guidance at all, and step 1 of the flow described it as **"the signed macOS app"**, which is the one thing it isn't.

## What changed

- A collapsed **first-launch note** under the download button with the working steps: launch once, then **System Settings → Privacy & Security → Open Anyway**. That path works on macOS 14, 15, and 26.
- It explicitly says Control-clicking **Open** does not work on macOS 15+, because that is the advice everyone finds elsewhere and it strands them.
- It says plainly that the dialog is about the missing Apple signature, not about anything found in the download, and points at the published SHA-256.
- Notes that Computer Use separately needs Screen Recording and Accessibility.
- Drops the "signed" claim from the Download step.

## This is interim

It comes out once the build is Developer ID signed and notarized — at that point none of it is true any more. The pipeline fix is Lore-Hex/QuillCode#new (`fix/notarize-macos-disk-image`); the remaining blocker is the Apple Developer ID certificate and notary key.

## Testing

`tests/test_public_revenue_pages.py` and `tests/test_public_seo_pages.py` — 101 passed. Rendered locally and checked both the collapsed and expanded states.

### DIFF
--- a/src/trusted_router/static/charter.css
+++ b/src/trusted_router/static/charter.css
@@ -875,6 +875,57 @@ tbody tr:hover {
   font-size: 12px;
 }
 
+.cc-first-launch {
+  max-width: 780px;
+  margin: -10px 0 24px;
+  border: 1px solid var(--border-default);
+  border-radius: var(--r-control);
+  background: var(--surface-card);
+  font-size: 14px;
+}
+
+.cc-first-launch summary {
+  padding: 12px 16px;
+  color: var(--text-display);
+  cursor: pointer;
+  font-weight: 500;
+  list-style: none;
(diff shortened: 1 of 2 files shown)


jperla:

PR #860, opened 2026-08-26
ci: bound the sharded test jobs

jperla:
One line per sharded matrix. A wedged shard currently burns GitHub's **6-hour default**; seen twice today — a `test-post-cutover` shard at **2h16m** (cancelled) and an earlier run's shard idle for hours, while every sibling finished under 9 minutes.

#844's `pytest-timeout=600` bounds a hung *test*. It does not cover collection, fixture teardown, or a wedged xdist worker — which is what these were. `timeout-minutes: 25` bounds the job whatever wedges, with ~3x headroom over the measured ~9m wall.

Does not diagnose the underlying wedge; it makes the wedge fail fast and visibly instead of silently eating a runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,12 @@ jobs:
     # so main silently stops reaching production -- 11 hours of it on
     # 2026-08-23, with every run green. This repository is public, so extra
     # standard runners cost nothing but concurrency.
+    # A wedged shard otherwise burns GitHub's 6-hour default: seen twice on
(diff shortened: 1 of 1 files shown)


jperla:
CI green on this head (949e40ca): https://github.com/Lore-Hex/quill-router/actions/runs/33010477667 — bounds each sharded test job at 25 minutes so a wedge fails fast instead of burning GitHub's 6-hour default. Two wedges observed 2026-08-26 (a shard at 2h16m, another run's shard idle for hours) while siblings finished under 9 minutes.

PR #862, opened 2026-08-26

jperla:
Increment C of the design in #843 (stacked on that branch — merge B first). Registered apps can add 0–300% markup on token costs through their delegated keys; the user pays base+markup, the owner earns 70% in TR credits, TR keeps 30%. Mirrors the Custom Models payout machinery exactly (same share constant, same idempotent movement row, same frozen-settle-body pattern) rather than inventing a parallel system. Hold and charge are marked up consistently so settle-side caps don't claw it back; markup terms freeze at authorize so nothing can reprice in-flight work. Writing the money tests surfaced a real defect: payouts were recorded under the custom-model movement kind in all three backends. Ten test groups, each verified to fail when reverted.


jperla:
Superseded by a clean re-application onto post-B main (B's squash-merge broke this branch's ancestry). Same reviewed implementation, no hand-merged conflicts in billing code.

PR #874, opened 2026-08-27

jperla:
Increment C of the design (A in #841, B in #843). Registered apps may add 0–300% markup on token costs through their delegated keys; the user pays base+markup, the app owner earns 70% of the markup in TR credits, TR keeps 30%. Mirrors the Custom Models payout machinery — same share constant, same idempotent movement row, same frozen-settle-body pattern — rather than building a parallel system.

**Three rules govern the money**, each pinned by a test that fails when reverted:
1. Hold and charge mark up the **same base** (including hosted-tool/media costs).
2. The **authorization decides the rate** (frozen at authorize); the **final charge decides the base** (derivation after every clamp).
3. The frozen outbox body is a **cross-check that can only reject** — it can neither create nor suppress a payout.

Those rules are the distilled output of five adversarial review rounds that found five real money defects: federated keys paying owners nothing on peer planes, hosted-tool costs held-but-not-charged, forgeable payout rows, omitted fields silently voiding a required payout, and a regional clamp making TR pay out more markup than it collected. Writing the money tests separately exposed payouts being recorded under the wrong movement kind in all three backends.

Re-applied onto post-B main because B's squash-merge broke this work's ancestry; the reviewed implementation was preserved and re-applied rather than hand-merged through conflicts in billing files. The three rules were re-verified by mutation afterward.

Known follow-up, not in this PR: the pre-existing user-model payout path has the same frozen-body trust weakness fixed here for app markup.


jperla:
Superseded by #NEW — that branch was built from a stale tree and deleted three other teams' merged PRs (including the authorize contention fix). Rebuilt additively on current main.

PR #891, opened 2026-08-27

jperla:
Increment C of the design (A in #841, B in #843). Registered apps may add 0–300% markup on token costs through their delegated keys; the user pays base+markup, the app owner earns 70% of the markup in TR credits, TR keeps 30%. Mirrors the Custom Models payout machinery — same share constant, same idempotent movement row, same frozen-settle-body pattern — rather than building a parallel system.

**Three rules govern the money**, each pinned by a test that fails when reverted:
1. Hold and charge mark up the **same base** (including hosted-tool/media costs).
2. The **authorization decides the rate** (frozen at authorize); the **final charge decides the base** (derivation after every clamp).
3. The frozen outbox body is a **cross-check that can only reject** — it can neither create nor suppress a payout.

Those rules are the distilled output of five adversarial review rounds that found five real money defects: federated keys paying owners nothing on peer planes, hosted-tool costs held-but-not-charged, forgeable payout rows, omitted fields silently voiding a required payout, and a regional clamp making TR pay out more markup than it collected. Writing the money tests separately exposed payouts recorded under the wrong movement kind in all three backends.

**Applied additively.** An earlier rebuild was caught in review deleting three other teams' merged PRs — including the per-key authorize contention fix that had just resolved a production incident. This branch deletes nothing: `git diff --diff-filter=D` against main is empty and `gateway.py` retains its admission control.

Full serial suite: **7512 passed, 0 failed**.

Known follow-up, not in this PR: the pre-existing user-model payout path has the same frozen-body trust weakness fixed here for app markup.


jperla:
Force-pushed a corrected branch. The prior push was destructive and review caught it: the commit carried the correct tree but was **parented on current main**, so it silently reverted the 13 PRs merged in between (including the authorize contention fix and the security-coverage CI gate).

My own guardrail check missed it because it compared against the tree's own base, where the diff looked clean — it was measuring the wrong thing.

Fixed by re-parenting the tree onto its true base and rebasing onto main (zero conflicts). Now verified against the correct comparison: `git diff origin/main...HEAD --diff-filter=D` is **empty**, 24 files touched, +1396/-45, and all seven other-team files are present and unmodified. The money implementation itself is byte-identical to what passed five review rounds.

PR #1014, opened 2026-09-01
Restart AWS analytics drain after code swap

jperla:
## Summary
- stop the active analytics drain before replacing its working directory
- restart it unconditionally after installing the new unit and environment
- verify the running process cwd resolves to the newly activated tree
- start every clickhouse-client child from / so a deleted inherited cwd cannot block delivery
- add regressions for deployment ordering and child cwd isolation

## Root cause
The installer moved /opt/tr-clickhouse to .previous while the active drain kept that directory as its cwd. systemctl enable --now did not restart an already-active service. On a later refresh, removing .previous deleted the live process cwd, causing every clickhouse-client insert to fail with cannot get current directory while systemd still reported active. The durable AWS outbox preserved all rows, but freshness lag reached about 47 hours until the unit restarted.

## Verification
- bash syntax check passed
- focused Ruff and mypy passed
- 78 focused AWS drain and writer tests passed
- 103 deploy-script execution tests passed

### DIFF
--- a/clickhouse/ingest_operational_outbox.py
+++ b/clickhouse/ingest_operational_outbox.py
@@ -444,6 +444,11 @@ class ClickHouseOperationalWriter:
                     command,
                     input=payload,
                     env=env,
+                    # Do not inherit the daemon's working directory. Deploys
+                    # replace /opt/tr-clickhouse atomically; if an old process
+                    # survives a bad rollout, that directory can be deleted and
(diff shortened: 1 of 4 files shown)


jperla:
Superseded by merged #1015, which includes the stop-before-swap/start-and-PID verification, subprocess cwd isolation, and wedged-drain exit behavior.

PR #1084, opened 2026-09-04
Implement PostgresStore.update_key so AWS and Azure reach parity

jperla:
`PostgresStore.update_key` raised `_not_implemented`, so on the AWS and Azure planes the OAuth authorized-apps surface (increment E) returned 501 for budget changes and revokes.

**Shared patch semantics.** `apply_key_patch` and `TYPED_LIMIT_PATCH_FIELDS` move out of the Spanner module into `storage_key_patch`, imported by both backends. Two backends applying a patch by two separate code paths is exactly how they drift — the field a new feature adds gets handled in one and silently ignored in the other. Spanner behaviour is unchanged. (Resolving the rebase caught this in the act: main's copy had a `scopes` branch, and I verified all eleven branches survived into the shared module before accepting the resolution.)

**One transaction.** The entity is read `FOR UPDATE`, and the entity write and the typed `tr_key_limit` write commit together, so a concurrent authorize cannot interleave and mint headroom against a stale cap. A metadata-only patch leaves the typed row untouched.

**A latent overspend bug, caught in review and confirmed against production.** Caps are *distributed* sub-budgets, not replicated: key `key_g9jW…` has `limit_microdollars = 30000000` on its entity and 16 shard rows whose `limit_micro` sum to exactly `30000000`. The first version wrote the *full* cap to every shard row — on a multi-shard key that authorizes `limit × shard_count`, i.e. $480 for a $30 key. Harmless today since Postgres only creates shard 0, so rather than invent an escrow algorithm this backend does not have, the assumption is now explicit and loud: more than one shard row raises, naming the key, **before** anything is written. Removing that guard fails the multi-shard test, which asserts neither row changed.

Full suite: **8664 passed, 0 failed**.

### DIFF
--- a/src/trusted_router/storage_gcp_keys.py
+++ b/src/trusted_router/storage_gcp_keys.py
@@ -14,7 +14,6 @@ import hashlib
 import uuid
 from typing import Any
 
-from trusted_router.money import dollars_to_microdollars
 from trusted_router.scopes import validate_api_key_scopes
 from trusted_router.security import (
     hash_api_key,
@@ -35,13 +34,13 @@ from trusted_router.storage_gcp_counters import (
     key_usage_shard_count,
 )
 from trusted_router.storage_gcp_io import SpannerIO, run_in_transaction_with_retry
+from trusted_router.storage_key_patch import TYPED_LIMIT_PATCH_FIELDS, apply_key_patch
 from trusted_router.storage_key_usage import api_key_from_json, api_key_usage_snapshot
 from trusted_router.storage_models import (
     ApiKey,
     ApiKeyUsageSnapshot,
     GatewayAuthorization,
     _is_byok,
-    iso_now,
 )
 from trusted_router.types import UsageType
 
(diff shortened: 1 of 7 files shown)


jperla:
Fable review on an isolated snapshot of `3f8e13da`: 128 tests in the update-key and synthetic-monitoring suites pass, ruff and mypy are clean. Mutations that disable the multi-shard fail-closed guard, retarget the typed-limit update to a nonexistent shard, or skip the typed-limit update entirely each turn `tests/test_postgres_update_key.py` red. Merging on green.

jperla:
Round 4 pushed (a468f78a). A 27-agent adversarial review (six dimensions, every finding refuted by three independent lenses) found seven defects in round 3, three blocking — including a Spanner regression the spec forbade: the hold freeze made every settle on an uncapped key write the hot api_key row. All seven fixed and mutation-verified; full suite 8789 passed, 0 failed. One pre-existing leak on the non-typed replay path is noted for follow-up.

PR #1090, opened 2026-09-05
Fix spend-lease cleanup and retired gateway monitoring

jperla:
## Summary
- Freeze an expired, locally drained lease in the same Spanner transaction that releases escrow; read predecessor ownership inside that transaction.
- Preserve all close guards and idempotency; refuse premature close and fail worker health for errors/dead records.
- Reconcile TR_REGIONS during synthetic image refresh so retired gateways cannot survive in stale job environment variables.

## Verification
The four initial regressions failed on the original code. Focused storage/worker tests pass, including concurrent transaction retry, exactly-once release, rollback, and a differential active-versus-frozen lifecycle test. Ruff and mypy pass. Full local pytest suite is running; CI must pass before merge.

## Operations
US/EU synthetic target lists have been corrected in production without changing secrets, identity, image, or history. After deploying the worker, requeue only the incident leases and verify reserved credit, predecessor counts, and repeated clean executions. Do not weaken billing guards or broaden ops IAM.


jperla:
Production repair verified with the existing deployment identity and no IAM changes. The scheduled worker is running ddec3943. Nine independently verified expired leases were closed through the normal guarded reconciler; affected credit shards released exactly the expected escrow without changing their credits or charged usage. Repeated subsequent passes report zero errors and zero dead rows. Both regional health jobs now exclude the retired gateway; no historical samples or alarm thresholds were changed.

Final local verification: ruff passes, mypy passes (349 source files), and full pytest passes with 8702 passed, 355 skipped, 10 xfailed. Four initial regression cases were verified to fail on the old code. The initial CI asset failures also reproduce on unchanged main; the generated Wafer manifest/card are refreshed in 4f5d328e. Waiting for the remaining CI shard before merge.

PR #1093, opened 2026-09-05
clickhouse: analytics outbox poller idle backoff and scan floor (Spanner CPU alert)

jperla:
Root cause of the "TR Spanner: high-priority CPU" alert firing since 2026-09-05 00:00 UTC.

The analytics outbox poller (`clickhouse/ingest_outbox.py`, running as `tr-clickhouse-ingest.service` on the ClickHouse primary) loops over the shards with a fixed 100 ms sleep and no idle backoff, and its two statements have no lower bound on `commit_ts`. The queue is almost always empty, about 6,600 rows an hour drained within seconds, yet it was polled roughly 64,000 times an hour, and the CPU of each poll grew from 12 ms to 18 ms over a day as Spanner accumulated deleted-row versions on this high-churn table under `ORDER BY commit_ts`. Together the two statements were the top consumers of Spanner CPU, about 1,400 CPU-seconds an hour.

- Idle backoff: a pass that reads nothing sleeps exponentially longer up to a configurable cap (default five seconds) and resets on the first row read, bounding the added delivery latency by that cap.
- Scan floor: the largest acknowledged commit timestamp per shard, held per process, bounds the scan so it skips the garbage; it is unset on start and after any error, so nothing is skipped, and the deletion of acknowledged rows remains the only durable cursor.
- One statement per idle pass instead of two.

Deployment is the VM pipeline path, not CI: the install script named in the PR is re-run against the ClickHouse primary after merge and the unit restarted. Fable watches Spanner high-priority CPU fall back under the 45% threshold afterwards.

Reviewed by Fable on an isolated snapshot with mutation verification, and independently root-caused by a parallel adversarial workflow (results in the review comment).


jperla:
Fable review on an isolated snapshot of `eb3287e7`: the analytics-outbox suite passes, ruff and mypy are clean. Mutations that make the floor exclusive, drop the floor reset on error, disable the idle-cap validation, or remove the per-pass pacing sleep each turn `tests/test_analytics_outbox.py` red.

An independent seven-agent production profile (run from the ClickHouse VM on the real replica) confirmed the diagnosis and the fix constraints: per-shard floors (never a global one with per-shard LIMITs), inclusive `>=`, advance only after the delete batch commits, reset on start and error, remove the per-pass lag probe, and pace every pass unless a full batch was drained. The poller was 72.6% of Spanner query CPU during a breach; the hourly trigger is one tenant's tool-choice job keeping every pass non-empty. Merging on green; the VM unit is redeployed by script afterwards.

jperla:
CI caught a canary boundary regression: check_scan_floors used <= while fetch intentionally includes commit_ts >= floor. Pushed the one-line correction to < without changing the test or any alarm thresholds. All 35 analytics outbox tests and ruff pass locally. Also handling the separate unique gateway trace index incident in #1097; production replacement index is prepared, old constraint remains until the new readers roll out.

jperla:
Full CI run 33979297065 is green, including coverage. Incident repair: deployed only clickhouse/ingest_outbox.py from c99599d8 through scoped IAP to tr-clickhouse-1, preserving the previous file, compiling before atomic replacement, and restarting only tr-clickhouse-ingest.service. Live SHA256 matches 00f7153d04af9a3a898b560ea33ce8974352d1e323c90af023188fdf63e09a67. No alarm thresholds, IAM, other workers, or balances changed. Holding the merge briefly so it does not supersede the in-progress #1097 billing deploy at its current-main gate; will merge once that gate passes and ensure deployment source remains consistent.

jperla:
Post-deploy evidence: worker started at 2026-09-05T17:07:04Z. Spanner QUERY_STATS_TOP_MINUTE shows the two SELECT shapes consuming 33.18 CPU-seconds at 17:06, then 0.197/0.189/0.177/0.178 CPU-seconds per minute at 17:12-17:15. This is query CPU, not a claim about total instance utilization. New floor-bounded fetch is live; initial drain recorded 24 rows delivered with zero insert errors. Full local suite also passed: 8,786 passed, 367 skipped, 10 xfailed; ruff/mypy green. Alarm thresholds remain unchanged.

PR #1097, opened 2026-09-05
Fix multi-call billing blocked by unique trace IDs

jperla:
## Root cause
Responses web search creates multiple independently billable authorizations under one parent audit trace. The Stage D unique request-ID index rejected the second settlement. Its durable retries then repeatedly failed and timed out the scheduled drain.

## Repair
Prepare a nonunique sparse trace index, deploy bounded readers that reject ambiguous evidence with 409, then explicitly retire only the obsolete unique index. True authorization/reservation/idempotency constraints and all billing formulas remain unchanged. Add a two-phase migration and incident runbook.

## Verification
The ambiguous-evidence regression failed on old code and passes on the fix. A real local Spanner emulator reproduces the original unique violation and verifies both settlements and replay after migration. Focused suite: 54 passed, plus six backend lookup cases and the emulator test. Ruff and mypy pass. Full local suite is running; required CI must pass before merge.

Production preparation is complete; the old unique index has NOT yet been removed. Retirement and normal outbox replay follow only after all readers are deployed.

### DIFF
--- a/docs/incidents/2026-09-05-shared-trace-settlement.md
+++ b/docs/incidents/2026-09-05-shared-trace-settlement.md
@@ -1,0 +1,54 @@
+# Shared trace ID blocked multi-call settlement
+
+## Cause
+
+The Stage D schema made `gateway_request_id` globally unique in the typed
+authorization table. It is a parent audit trace, not a billing idempotency key.
+Responses web search, fusion, and other multi-call requests can legitimately
+settle several different authorizations under the same trace. The first
+settlement succeeded; the next rolled back at the unique index. Background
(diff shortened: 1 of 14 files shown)


jperla:
Full local regression suite completed: 8,783 passed, 368 skipped, 10 xfailed; the six additional Memory/Postgres lookup cases passed separately. Real Spanner emulator regression reproduced the original unique-index failure and proved multi-call settlement plus no-op replay after retirement. Production deployment 33979631020 has four ready new revisions at no traffic so far; old unique index is still retained until regional reader promotion completes. CPU pressure was independently reduced by #1093 and the follow-up operational pacer is #1098.

PR #1105, opened 2026-09-05
Improve chat UX, full model names, and bottom-right usage totals

jperla:
## Summary
- Keep chat inside the viewport with a roomier composer, quiet icon toolbar, simpler empty state, and readable model names in the picker and selected-model row.
- Show dollar cost and tokens at bottom right; expandable input, cached-input, output, and reasoning details. Read authoritative root usage (including zero costs), preserve cumulative snapshots, and mark missing billing as unreported instead of estimating a charge.
- Request streaming usage, show thinking before completion, and restore Send after upstream failures.
- Fix first-visit model deep links crashing before state initialization and refresh model labels after the catalog arrives.
- Reuse the existing chat path for user-chat. No backend billing, provider routing, auth, or enclave changes. Totals explicitly cover current responses, excluding replaced/deleted branches.

## Verification
- Full Python suite: 8883 passed, 368 skipped, 10 xfailed.
- ruff and mypy passed; chat CSS lint passed.
- Full website browser suite: 40 passed (11 new chat regressions).
- Desktop 1280px and mobile 390px/320px screenshots: model names wrap, no horizontal overflow, composer and usage stay visible.
- Mocked streaming covers real-time reasoning/usage, root costs, duplicate cumulative chunks, partial usage events, free/missing/invalid costs, persistence, clearing, signed-out gating, failure cleanup, and custom-model locking.
- Local Claude CLI Opus review attempted but unavailable: OAuth session expired. Manual review completed.

## Rollout
Standard guarded control-plane deployment on merge. Follow with live page and tiny streaming usage smoke.


jperla:
The live chat-proxy smoke exposed an existing direct-stream omission: token usage was emitted, but settled billing cost was not. Companion fix: https://github.com/Lore-Hex/quill-cloud-proxy/pull/313. It keeps answer/thinking streaming and adds one authoritative usage record before DONE, using the existing settlement path. The UI deliberately displays Cost unreported until a real billed amount is available.

PR #1108, opened 2026-09-06
Schedule NEAR AI GLM 5.1/5.2 retirement for September 11

jperla:
## Change
- Retire only NEAR AI `zai-org/GLM-5.1-FP8` and `z-ai/glm-5.2` at **2026-09-11 13:00 UTC** using the shared lifecycle policy.
- Keep the original upstream IDs until cutoff; do not silently substitute GLM 5.3 Flash or affect equivalent models on other providers.
- Filter retired workloads consistently from authenticated pricing discovery, expected-model validation, missing-price checks, and coverage normalization. Active workloads still fail closed on missing prices or attestation/domain drift.
- Keep the GLM 5.3 Flash replacement behind its own attestation review. The live NEAR direct registry publishes it, but our released workload allowlist does not yet pin it.

## Validation
- Reproduced **6 failing regressions before the fix**.
- **20 focused tests passed** with the normal clock and **20 passed** with the post-cutover clock (2027-01-01).
- Tests cover exact cutoff, provider scope, running-server filtering without catalog reload, unchanged upstream IDs, stale feeds, missing pricing, and replacement not bypassing attestation.
- Whole-tree Ruff and mypy pass. Full local suite/coverage running; CI required before merge.
- Claude CLI Opus review attempted, but local CLI reports not logged in. No API-key fallback was used.

## Rollout
Normal CI-gated control-plane deployment. No enclave changes, credential changes, or immediate traffic migration.

### DIFF
--- a/scripts/pricing/providers/near_ai.py
+++ b/scripts/pricing/providers/near_ai.py
@@ -25,6 +25,7 @@ from scripts.pricing.base import (
 )
 from scripts.pricing.manifest import write_discovered_chat_manifest
 from scripts.pricing.model_ids import remember_upstream_id
+from trusted_router.provider_lifecycle import provider_model_retired
 
 SLUG = "near-ai"
 CATALOG_URL = "https://cloud-api.near.ai/v1/models"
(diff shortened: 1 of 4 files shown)


jperla:
Read-only live validation: the direct endpoint registry publishes `z-ai/glm-5.3-flash` at `glm-5-3-flash.completions.near.ai`, and the authenticated catalog includes pricing for both retiring GLMs and the proposed replacement. Full NEAR refresh currently fails closed because these unrelated pinned direct endpoints have no corresponding authenticated pricing row: `Qwen/Qwen3.5-122B-A10B`, `Qwen/Qwen3.6-27B-FP8`, `google/gemma-4-31B-it`, and `openai/gpt-oss-120b`. This PR deliberately preserves that guard and does not publish stale or guessed prices. Scheduled runtime retirement remains independent of successful refresh. No production data or routing was mutated during validation.

jperla:
Validation update: all six regular test shards and all six post-cutover shards passed in CI. Local full-suite run was stopped after 2,792 passes to diagnose one unrelated Hypothesis timing flake in unchanged test_v2_still_rejects_a_wrong_binding (334ms vs 200ms deadline, then 130ms on retry). The exact test passed immediately in isolation with coverage still enabled; no deadline or BYOK code was changed. Local full-suite completion is not claimed. Focused normal/future runs, whole-tree Ruff, and mypy all passed. Awaiting combined CI coverage gate before merge.

PR #1110, opened 2026-09-06
Prevent full-table Spanner payment ownership searches

jperla:
## Incident
Three administrative Stripe-ID substring searches scanned the legacy entity table and consumed about 91 CPU-seconds. The strict 45% CPU alert correctly fired and recovered. The original caller is not proven by the available audit evidence. Customer identifiers are intentionally omitted.

## Fix
- Add a read-only PaymentIntent ownership helper: one exact Stripe GET, at most two complete-key Spanner reads, LOW priority, five-second RPC deadlines, no retries or arbitrary SQL.
- Use the generated Spanner RPC client without a background metrics exporter or session pool.
- Require opt-in email output and distinguish the current workspace owner from a recorded initiating user.
- Document safe incident queries in AGENTS.md, CLAUDE.md, the runbook, and an incident report.

## Verification
- Ruff and mypy passed, including the new script.
- 24 focused tests passed; the full local suite passed with 8,976 tests, 368 skips, 10 expected failures, and 84% coverage.
- Exact read-only production lookup succeeded for all three searched PaymentIntents.

No balances, IAM, application deployment, or alert thresholds changed. This is an approved safe lookup path, not an IAM-enforced SQL firewall.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,6 +51,17 @@ is a no-op — the next rollout overwrites it.
 
 **Never** run large unbatched DML against production Spanner during a rolling deploy.
 
+**Read-only is not load-safe.** Never search production `tr_entities.body` with
+`LIKE`, regular expressions, or JSON predicates without a complete indexed key.
+`LIMIT 5` limits results, not the rows scanned. Do not discover entity kinds with
+wildcard `COUNT(*)` queries. Three ad hoc Stripe-ID body searches caused the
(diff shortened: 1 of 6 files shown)


jperla:
Final local verification: 8,976 passed, 368 skipped, 10 expected failures; coverage 84%. Ruff and mypy passed. The bounded read-only production lookup passed for all three investigated payments. No application rollout or alert-policy change is required.

PR #1125, opened 2026-09-07
Fix Featherless cached-input pricing discovery and billing

jperla:
## Summary
- Import Featherless cached-input pricing automatically during the existing hourly refresh. Its authenticated `/v1/models` currently omits cache prices; the official request-pricing tables publish them.
- Parse named columns and prefer exact model overrides over class rates. Preserve authenticated input/output prices; require them to match the public rate before importing a class/model cache discount. Explicit API cache prices take precedence, including zero.
- Fail closed on malformed/conflicting pricing or missing table structure. Unknown cache prices stay unknown, never zero. Keep the existing shared manifest, customer markup, cache-usage normalization, and settlement logic.
- Refresh all 17 existing Featherless routes with exact published cache rates. No input/output price changes or new routes.
- Regenerate Xiaomi's social card after the already-scheduled UltraSpeed retirement. Its stale counts failed identically on unchanged main; no retirement behavior or test gate was changed.

## Evidence
Official source: https://featherless.ai/docs/request-pricing-and-credits

Selected upstream cached input rates per million: GLM 5.3 $0.26, GLM 5.3 Flash $0.03, GLM 5.2 $0.15, Kimi K3 $0.30, DeepSeek V4 Flash 0731 $0.03. Published customer rates retain the normal markup.

Direct live GLM 5.2 repeat-prefix smoke: HTTP 200; one response reported 3,701 prompt tokens and 3,701 cached tokens in 1.35 seconds, versus the first uncached request in 8.59 seconds. Other calls missed; cache hits are not guaranteed. The gateway already normalizes Featherless's top-level `usage.cached_tokens`, so no enclave change is required. GLM 5.3 Flash probes reported zero cached tokens; no cache-hit claim is made for those probes.

## Verification
- Regression failed before implementation.
- 22 new parser tests pass; focused gateway settlement and catalog tests pass. The money-unit guard rejects missing per-token fields instead of falling through to Featherless's per-million aliases.
- Shared settlement regression now covers Featherless cache-subset billing, recorded cached usage, and duplicate-settlement idempotency.
- `ruff check .` and `mypy` pass. Full local suite: 9,665 passed, 408 skipped, 11 xfailed. Local fake-server tests were run with loopback networking permitted. Full CI required before merge.
- Local Claude CLI Opus review attempted, but CLI reports not logged in. No private API-key fallback used.


jperla:
Additional verification: the full local suite completed with 9,638 passing tests. Nineteen socket tests failed under the sandbox and all passed with local networking permitted. The only remaining two failures reproduce on unchanged main: the scheduled Xiaomi UltraSpeed retirement has reduced the live route/model counts. The follow-up commit regenerates only its stale card/manifest; no routing behavior or test safety gate is changed. Focused pricing, settlement, and branding tests now pass. Full local and CI suites are rerunning on the final head.

PR #1136, opened 2026-09-08
Fix Sentry trust-verdict expiry, playground stream failures, and stale Akash route

jperla:
## Summary

Repair three independently verified failure paths:

* Refresh successful global billing-trust verdicts with five seconds of headroom before entering regional transactions. Keep the original 15-second validity ceiling, marker deadlines, cached refusal TTL, and single-flight refresh. No transaction-time fleet reads or billing enforcement changes.
* Make the playground proxy return 502 on an upstream failure before its first byte, emit an explicit SSE error on an interrupted uncompressed stream, preserve opaque compression headers, and never retry an already-sent inference request. Record request/workspace/credential IDs, outcome, and byte counts without bodies or raw error text.
* Make the browser recognize SSE errors and premature EOF instead of silently accepting partial answers. Disable only Akash's confirmed missing DeepSeek V4 Flash 0731 route and remove that missing model from refresh requirements. Other Akash routes and other providers of DeepSeek remain available.

## Evidence

New trust and proxy regression tests were run on the old implementation first and failed. Browser error-event and truncated-stream tests also failed on old code and now pass.

Akash's authenticated current model list excludes the native DeepSeek ID. Two direct tiny streaming probes returned HTTP 200 with an SSE `model_not_found` error and no tokens. The route is marked canary-failed so it must pass a future provider canary before restoration.

## Validation

* Ruff and mypy pass.
* 132 trust tests pass.
* 12 proxy tests pass.
* 108 provider/catalog coverage tests pass.
* All 13 local chat Playwright tests pass, using mocked providers.
* TypeScript check, ESLint, and Stylelint pass.
* Full four-worker Python suite and coverage check are running; results will be added before merge.

No alert thresholds, privacy scrubbers, credentials, or production balances changed. This repair does not claim to replenish upstream provider credit or prevent every network interruption.


jperla:
Final local acceptance run: 9,937 passed, 408 skipped, 11 expected failures; coverage 84.36% (70% gate). Ruff, mypy, TypeScript, ESLint, Stylelint, 13 playground browser tests, and targeted regression suites passed. Provider card artifact was regenerated after disabling the missing Akash route, and branding tests passed. Gateway root cause for the incomplete stream is fixed separately in Lore-Hex/quill-cloud-proxy#321: missing terminal HTTP chunk after an explicit SSE provider failure. Its regression failed before the fix for both Chat and Responses with keep-alive enabled; full production-tagged Go suite with race detection passed afterward. No alert thresholds or trust evidence lifetimes were weakened.

PR #1166, opened 2026-09-12
Keep provider failures separate from rollout health

jperla:
## Summary
- Preserve gateway error.source in SDK synthetic probes as provider_error or router_error. Keep HTTP status, and never export the error body.
- Scope legacy watchdog payloads to the requested router-core or control-plane probe families, matching the SLO-aware path.
- Keep provider catalog health independent of release CI and document release ordering with the companion enclave change.

## Verification
- Regression-first: 26 new cases failed on the previous implementation; all 76 focused tests pass after the fix.
- Ruff and mypy pass.
- Full release suite is running with local socket permissions; provider_health remains its separately reported operational workflow.
- No production failure injection or new provider calls.

Deploy this attribution producer before the quill-cloud-proxy gate consumer.

### DIFF
--- a/docs/provider-independent-deploys.md
+++ b/docs/provider-independent-deploys.md
@@ -1,0 +1,39 @@
+# Provider Health And Deployment Gates
+
+Upstream model availability is a separate operational signal from release
+correctness. A provider outage must not be treated as a broken router release.
+
+## Control Plane
+
+- CI runs offline routing, expiry rejection, billing, fallback and privacy tests.
+- Time-sensitive catalog counts run in `Provider catalog health`, independently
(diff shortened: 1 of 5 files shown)


jperla:
Local release verification complete: Ruff passes; mypy passes (375 source files); 10,376 tests passed, 408 skipped, 11 xfailed with the release marker selection (provider_health remains separate). The 26 new regression cases were run against the old implementation and failed first; all 76 focused tests now pass. Broader companion rollout tools: 512 passed, 1 skipped, 109 subtests; cloud_gcp,llm_multi Go tests pass. No live error injection or paid provider calls were added.

PR #1179, opened 2026-09-14
Create Lightning accounts only after payment

jperla:
## Change
Unpaid invoices now create only pending checkout metadata. No TrustedRouter user, workspace, API key or grant is provisioned until LND settlement is verified and stored. The account/key UI waits for USD credit acknowledgement. Existing-key top-ups retain normal key revocation checks.

Pending recovery credentials use AES-GCM with a domain-separated key and authenticated checkout fingerprint. Ciphertext is cleared once the paid identity is bound. This preserves background recovery and exactly-once credit retries without plaintext keys in the database.

The funding database is not deployed yet; its first production schema now uses lr_checkouts. Mainnet payments remain disabled.

## Verification
The new unpaid-account regression failed against the previous implementation. Local Lightning tests: 151 passed across SQLite/PostgreSQL, 88.51% coverage; six browser smokes and 20 JavaScript tests passed. Experiment and root ruff/mypy passed. Full root pytest running; CI must pass before merge.


jperla:
Final validation: every PR check passed on dbd14143e8e1ea71e1cc0cd271c4662ae31e83ed, including full repository CI and the dedicated Lightning workflow. Local Lightning suite: 151 passed (SQLite/PostgreSQL), 88.51% coverage; 20 JS tests and 6 browser tests passed. Local root lint and mypy passed. The full local root run completed with 10,645 passed, 471 skipped, 11 xfailed and one provider-image catalog-count mismatch in tests/test_provider_branding.py::test_provider_social_card_counts_match_current_catalog. That mismatch reproduces in isolation; src/ and that test are identical to origin/main. It is outside this payment-only change and was not hidden or modified. No mainnet funding enablement or production database migration is part of this PR.

PR #1188, opened 2026-09-15
Fix Lightning activation for gcloud singleton responses

jperla:
Production preflight found gcloud returning a singleton list for a global Cloud Armor policy. Normalize that documented CLI result shape without accepting ambiguous or wrong-policy results. Keep both invoice and general ingress rate limits unchanged. Refresh the Lightning README to point to live readiness and the production runbook instead of stale syncing information. Six regression cases failed on the old implementation and pass after this change. Ruff, mypy, and all 15 activation tests pass; full suite is running. No application image, money logic, or TrustedRouter release gates are changed.


jperla:
Complete local verification passed: 10,795 tests passed, 471 skipped, 11 expected failures. Ruff and mypy are clean. The real Cloud Armor response reproduced the singleton-list regression; both rate limits are now attached successfully to the isolated Lightning backend, with CDN disabled. No runtime image or billing logic changed.

PR #1194, opened 2026-09-15
Harden Lightning credit delivery and recovery

jperla:
## Summary
Fix invoice rounding and receipt caps, preserve overpayments, and add durable credit-review states, retry backoff, stuck-payment health and heartbeat alerts. Harden missing-invoice recovery, node-index reuse, secret rotation, rate limiting and bounded unpaid-record retention.

## Tests
Final root run: 10817 passed, 471 skipped, 11 xfailed. Funding SQLite/PostgreSQL: 240 passed with 89% coverage, including migrations and two-process settlement. Ruff/mypy clean; 20 JS tests and all 7 browser tests pass. Initial regressions and four Fable follow-up regressions failed the previous implementations. Full Fable review plus independently checked follow-up fixes.

## Release
Normal CI and control-plane deployment first, then immutable funding image/migration, alerts and node invoice-only maintenance. No paid receipt is pruned. No outgoing payment, live wallet restore, or simulated production settlement. Real paid website smoke and isolated restore drill remain separate acceptance checks.


jperla:
Fable follow-up reviewed the staged source. I reproduced four additional failures on the prior snapshot, then fixed and tested them: fresh quote cache readers contending on the fetch lock, resumed-orphan cleanup interruption, cancellation delayed by review backoff, and a formatter-independent alert trigger. Funding SQLite/PostgreSQL is now 240 passing tests at 89% coverage; all 7 browser tests pass. Alert installation also waits for an actual post-metric heartbeat. The alleged orphaned-paid-invoice loss was not reproducible because the existing FK and account-row check prevent invoice creation after deletion; the checkout interruption was real and is fixed. Built JS synchronization and real LND LookupInvoice NOT_FOUND handling are verified. No production payment or restore was performed.

PR #1195, opened 2026-09-15
Add LightningRouter reasoning effort and recover invoice cancellation

jperla:
## Changes
- Add Default, Low, Medium and High reasoning effort to the client setup snippets, gated by advertised model capability.
- Generate the native OpenCode, Crush and OMP effort settings without changing funding or inference routing.
- Recover a lost cancel response with a bounded status lookup in both the LND adapter and browser. Never replace an unresolved invoice or discard its original key.
- Preserve exactly-once credit when payment wins cancellation.

## Verification
- New adapter and browser cancellation regressions failed against the previous implementation and pass with these fixes.
- Full local funding suite on SQLite/PostgreSQL: 254 passed at 89% coverage before final extra adapter cases; final rerun in progress.
- JavaScript: 29 passed. Browser: 11 passed across mobile/tablet/desktop.
- Root and funding mypy passed; root ruff passed. Full repository pytest is running.
- No outgoing Lightning payment, wallet change or node restart. Deploy only the isolated funding web image after CI passes.


jperla:
Final local gates passed: root ruff and mypy; full repository pytest 10,817 passed, 471 skipped, 11 xfailed; funding mypy and SQLite/PostgreSQL pytest 259 passed at 89.60% coverage; JavaScript 29 passed; Playwright 11 passed. The original adapter cancellation regressions failed in 7 cases; the original lost-response browser regression failed before implementation, and all now pass. Mobile 375px screenshot inspected for layout. Deployment remains pending required CI and refresh of the separate deployment account credentials; the ops identity is unchanged.

PR #1201, opened 2026-09-15
LightningRouter community links and BSL source repository

jperla:
## Changes
- Link LightningRouter and TrustedRouter on X, the existing Discord invite, source code, and GitHub Issues from every LightningRouter page.
- Publish source-available BSL 1.1 licensing, preserve third-party notices, and distinguish private payment/security help from public support.
- Add a deterministic committed-source exporter and a tested-source mirror workflow. Only allowlisted LightningRouter files are published; no production deployment credentials are needed by the mirror.
- Configure the new Lore-Hex/lightning-router repository with Issues enabled and pull requests disabled.

## Verification
- Root ruff and mypy passed. Full root pytest is running.
- LightningRouter ruff/mypy, 266 Python tests, 120 JavaScript tests and 47 browser tests passed.
- Six export regression tests passed; dirty/untracked files, database/key paths and symlinks are excluded or rejected.
- Mobile and desktop page screenshots checked; all footer links fit with 40px touch targets.

No billing, custody, FX or invoice state-machine changes.

PR #1203, opened 2026-09-15
Include inherited esbuild config in LightningRouter source export

jperla:
The newly published standalone repository caught an inherited build configuration missing from the export: esbuild reads the monorepo root tsconfig.json, and its strict setting changes the generated block-scoped function output. The source repository CI passed, but the exported repository failed its committed-bundle parity check.

Include the existing committed root tsconfig.json in the allowlist and assert its byte-for-byte export in the regression test. No application, billing or deployment behavior changes. Rebuilt both bundles in the standalone export with the config present and confirmed both exactly match the committed bundles.

Root ruff/mypy and focused exporter tests pass. Full root pytest is running again. Follow-up to #1201.


jperla:
Root cause confirmed locally: adding the committed root tsconfig.json makes both exported browser bundles byte-identical to source. CI now rebuilds a fresh committed export in RUNNER_TEMP and cmp-checks both bundles before allowing release. An unrelated CI teardown race was traced to async route.fetch handlers running while the browser context closed; the shared page fixture now drains handlers with unrouteAll({behavior: wait}) before disposal, without ignoring errors. All 47 browser cases passed three consecutive local runs (141 tests). Website code is unchanged and #1201 is already deployed and production-smoked successfully.

jperla:
Release-gate investigation: the post-cutover sitemap failure also occurs on current main (run 35016296670). It is deterministic test pollution, not a production/source-export regression. With TR_LIFECYCLE_CLOCK_OVERRIDE=2026-10-10T16:00:00Z, running the MiMo alias GET test before the sitemap hygiene test failed: a temporary model survives in the comparison LRU after monkeypatch restores the catalog. Running sitemap alone passes. Commit a8e4d349 clears the comparison cache around the two catalog-mutating alias tests. The exact formerly failing two-test sequence now passes. No application code or assertions weakened. Root ruff and mypy pass. The complete local root suite passed 10,824 tests before this fixture-only correction; the full suite is running again with the correction. The fresh-export bundle parity gate and 141 repeated browser tests already passed.

jperla:
Final local validation after the cache-isolation correction: uv run ruff check . and uv run mypy passed; uv run pytest -q -n 4 --dist loadgroup completed with 10,824 passed, 471 skipped, 11 xfailed (5m03s). The formerly failing post-cutover CI shard is now green, and LightningRouter CI including its standalone export-build comparison passed. Remaining standard root CI shards are still running. Production community links are already deployed and verified in #1201.

jperla:
While the original head completed all CI gates, #1204 merged an independent fix for the same SEO test pollution. Merge f8cfb015 retains #1204 exactly (including its derived-index cleanup) and removes the duplicate fixture from this branch. The PR is back to four LightningRouter-only packaging/CI/test files, with no application behavior changes. Current main is incorporated without rewriting history; normal CI and the local full suite are running on the exact merged tree.

PR #1204, opened 2026-09-15
Fix SEO alias-test cache pollution blocking guarded deploys

jperla:
## Root cause
Under the post-retirement lifecycle clock, the MiMo alias test inserts a retired model into MODELS and renders a page that populates three cached comparison indexes. Monkeypatch restores MODELS but not the derived caches. The later full-sitemap test receives a stale MiMo comparison URL and correctly rejects its 404. Whether it fails depends on worker test order.

## Fix
Clear the three derived comparison caches before and after the tests that temporarily insert or remove catalog entries. No production code, sitemap assertions, privacy rules or deployment gates change.

## Verification
- Exact ordered two-test reproduction failed before the fix and passed afterward with TR_LIFECYCLE_CLOCK_OVERRIDE=2026-10-10T16:00:00Z.
- Ruff and mypy pass.
- Full local suite and required CI are running.
- Local Claude CLI Opus review requested.

Unblocks the Telnyx metadata release in #1202 without weakening its release gates.


jperla:
Full local suite passed with localhost networking permitted: 10,828 passed, 471 skipped, 11 xfailed. Ruff and mypy also pass. The exact future-clock alias-then-sitemap sequence failed before the change and passed after it. The original runtime commit CI passed its single retry; this permanent test-only cleanup remains necessary to remove the proven order dependence.

jperla:
Claude CLI Opus completed a static review: the three-cache cleanup is correct, teardown ordering is safe, and no assertion was weakened. Keeping setup cleanup intentionally makes each catalog-mutating test independent of warm-cache order. A global autouse reset would penalize every test and is outside this narrow release unblock; broader catalog-fixture cleanup can be handled separately. The reported two-test reproduction and full-suite results were run directly here, not claimed as executed by the reviewer.

PR #1213, opened 2026-09-16
Fix credit rebalance cooldown races and direct DeepSeek inventory

jperla:
## Changes
- Recheck a concurrent credit rebalance after at most two 250 ms waits outside transactions. Preserve the cooldown, three-attempt ceiling, aggregate affordability checks, idempotency, and the existing request deadline.
- Log bounded cooldown exhaustion with workspace metadata only.
- Make the authenticated DeepSeek model manifest authoritative for both prepaid and BYOK, keeping unsupported shared-catalog IDs out of direct routing while preserving independently verified providers.

## Verification
- Four billing regressions and four catalog regressions failed before the implementation change.
- 79 targeted tests pass, including live-expansion conservation and replay tests.
- Ruff and mypy pass. Full local suite is running; merge is held for complete CI and local validation.
- Cost-free authenticated DeepSeek inventory confirms deepseek-flash and deepseek-v4-pro. No raw credentials or request content are logged.

No pricing, balances, shard counts, instance capacity, alert thresholds, or attestation policies are changed.


jperla:
Final local verification completed: 10,875 passed, 471 skipped, 11 expected failures. The full coverage run reached 84.50%; its sole stale generated-asset failure was corrected by regenerating only the changed DeepSeek card, followed by the successful full rerun. Ruff and mypy passed. Fable review found no concrete blocking defect. The new billing and catalog regressions were confirmed failing on the old implementation.

PR #1215, opened 2026-09-16
Shorten price refresh validation without dropping any gates

jperla:
The last successful refresh spent 34m52s in its serial full-test prepublication step before running exact-snapshot CI again. That leaves little time in the hourly refresh window and increases the chance main advances before publication.

Use exactly the existing CI command: pytest -q -n 4 --dist loadgroup. This preserves the entire suite and its file-group isolation. Lint, mypy, spike guard, coverage audit, exact-SHA snapshot CI, and stale-main rejection are unchanged.

Verification: the exact grouped-worker full suite passed locally (10,903 tests in 5m17s), workflow regression checks pass, ruff/mypy pass. No runtime code changes.


jperla:
Merged after CI 35075309637 passed and after refresh 35072409935 successfully published snapshot 0649a55ac1a93b7bd3ea7baaead023463ec555ca. The previous serial refresh validation took 36m03s for 10,903 passing tests. The new workflow uses the same four-worker grouped full-suite command as CI, without removing lint, types, pricing spike guard, catalog audit, exact-snapshot CI, or stale-main fences. Local ruff, mypy, and focused workflow tests passed; the prior full local suite also passed 10,903 tests.

PR #1222, opened 2026-09-16
Repair BTCPay node health with private NBXplorer

jperla:
## Problem
The BTCPay dashboard connected the existing LND node but had no NBXplorer. Its default explorer URL pointed at an empty loopback port, producing the offline/sync warning despite a fully synced funded node.

## Repair
- Add digest-pinned NBXplorer on the isolated dashboard VM, with its own database and authenticated internal API.
- Reuse existing Bitcoin Core and LND; preserve wallet, channel, chain data, existing RPC auth and checkout.
- Restrict new Core RPC/P2P access to the dashboard private IP and an explicit non-spending RPC allowlist.
- Require backup, no in-flight HTLCs, identity preservation, rollback on failed node restart, and real indexer/dashboard synchronization gates.
- Back up both dashboard databases and preserve the private configuration on future node installs.

## Verification
- 26 focused tests pass; ruff and mypy pass.
- Real local Bitcoin 31.1 + NBXplorer 2.6.14 + BTCPay 2.4.4 integration reports synchronized=true.
- Unauthenticated indexer requests return 401; node stop RPC with the indexer credential returns 403.
- Full repository suite and CI in progress.

Production activation uses the committed manual explorer deployment, not a replacement node or wallet.


jperla:
Final local verification on 796a2a28: 10,969 passed, 471 skipped, 11 expected failures; all 26 BTCPay tests, Ruff, and Mypy pass. The real isolated regtest stack passed authenticated synchronization, unauthorized API rejection, forbidden stop RPC, dependency outage/recovery, and an NBXplorer backup/restore with 25 matching public tables. Authenticated production dashboard reproduction confirmed the sync warning before deployment.

Operational rollout will also refresh the existing VM bitcoin-bootstrap metadata from the committed bitcoin_node.py change. Read-only comparison confirmed the currently deployed bootstrap exactly matches pre-repair main, so this preserves existing customization while preventing reboot from reverting the private listener. No second mainnet node, wallet replacement, or funds movement.

jperla:
Deployed and verified in production. BTCPay public health now returns synchronized=true; authenticated dashboard no longer renders syncModal. Existing Bitcoin/LND identity was preserved; both chain and graph are synced, one funded channel is active, zero pending HTLCs. LightningRouter /health remains up with payments_ready=true and zero uncredited deliveries. Existing LND backup and both BTCPay/NBXplorer database dumps succeeded. Production scoped RPC can read getblockchaininfo but getrpcinfo and sendtoaddress return HTTP 403. Bitcoin listeners are loopback/private only, and ingress is limited to the dashboard private address. Greg remains Observer, signup remains disabled, and no second mainnet node or wallet was created. The saved Bitcoin VM bootstrap was updated and hash-verified to preserve this configuration on reboot.

PR #1227, opened 2026-09-17
Bound recovery of expired unpublished Lexe invoices

jperla:
Follow-up to #1226 before production activation. An authoritative empty recovery scan may close only an entirely unbound, unpublished intent after quote expiry plus 120 seconds. Concurrent binding, provider outage, duplicate matches and truncated scans remain fail-closed. Preserve a terminal audit code without wedging checkout health. PostgreSQL and SQLite regressions cover expiry, outage, stale failure and settlement races. Actual receive-only Lexe node recovered a fresh pending invoice and canceled it; no funds moved.


jperla:
Deployed and verified in production. Final image sha256:3db9a0f209adfff721800a84ef66d96fc591c95838957c10faa7c3a0e8a39c02; revision lightning-router-web-lexe42d5cfd1-live serves 100%, new invoices use Lexe. Normal CI gates passed. Local root suite: 11,076 passed; funding suite: 670 passed across PostgreSQL/SQLite, 91% coverage. Fable found no blockers in the recovery fix. Staged/warmed the dual-backend bridge, passed public LND create/cancel, drained and retired incompatible web revisions, then warmed/promoted Lexe. Public Lexe invoice matched the authenticated wallet and canceled successfully, including repeat cancellation. Health is up, payments_ready true, uncredited_count=0, review_required=0. No funds moved and no customer account was created by these unpaid canaries. Funded end-to-end receipt and USD crediting remain to be verified by Greg paying through the unchanged website. The 10% FX buffer and USD balances are unchanged. Existing Bitcoin/LND/BTCPay nodes, funds and backups remain intact; rollback must use the dual-backend image.

PR #1229, opened 2026-09-17
Pad LightningRouter setup tab highlights

jperla:
## Summary
- Add 16px horizontal padding and reduce external spacing between agent tabs.
- Keep a balanced two-column tab layout on narrow phones.
- Cover all four tabs in light/dark mode at six mobile and desktop widths.

## Verification
- New regression failed on the original CSS (0px padding), then passed after the fix.
- Visual review of desktop and mobile screenshots passed.
- Full local suites and CI are running; deployment will preserve the live Lexe backend and wallet settings.

No payment, invoice, key, or provider logic changes.


jperla:
Validation complete: the new regression failed against the original 0px horizontal padding, then passed after the fix. All 77 browser tests passed, including both themes at six viewport widths. Root lint, type checks, and the full test suite passed (11,076 tests). The working tree is clean.

PR #1230, opened 2026-09-17
Document confidential-only API origins and explicit privacy setting

jperla:
Documents the new confidential-only origins, required provider.min_privacy=confidential setting, supported routes, fail-closed behavior, and hardware-attestation verification caveat. Companion enclave PR: https://github.com/Lore-Hex/quill-cloud-proxy/pull/347. HOLD publication until all four names pass live certificate, attestation and inference-policy checks. Validation: ruff, mypy, 11076 full-suite tests passed; two new documentation tests passed separately.


jperla:
Hold publication: implementation merged as quill-cloud-proxy a2e9659 and CI is green; regional rollout 35172730577 is running. New-name certificate bootstrap is currently blocked: Cloud DNS TXT changes from the GCP enclave workload return 403 Forbidden (quill-workload service account). Ordinary TLS and API health work. DNS membership requires all four actual confidential-SNI certificates plus live attestation/policy, so new A records remain absent. Do not merge these docs until that blocker is resolved and all four hostnames pass the live contract; no workload IAM privileges were added.

jperla:
Additional blocker detail: the signed-in owner can read the zone policy, but the automated security reviewer rejected the proposed conditional DNS role grant. No IAM mutation occurred (etag and bindings verified unchanged). Explicit owner approval is required before any grant attempt. Do not bypass this through another identity or workflow. Existing API health and the new binary Host-policy attestation probe passed; actual confidential-SNI TLS remains unavailable.

PR #1232, opened 2026-09-17
Fix Lexe invoice timestamp boundary validation

jperla:
## Cause
A real unpaid invoice was signed one second before Lexe persisted its payment record. The adapter treated those independent timestamps as identical and quarantined the invoice, leaving the health endpoint degraded and repeating payment-review alerts. Lexe assigns created_at on first persist after signing.

## Fix
Validate timestamp freshness and ordering instead of equality. Preserve signature, wallet, amount, hash, expiry, and settlement-proof checks. Include only allowlisted failure codes in reconciler logs. Existing quarantined rows recover through the normal idempotent worker, not manual balance edits.

## Verification
Five new regression cases failed on the original adapter before the fix. Funding suite: 688 passed on SQLite/PostgreSQL with 90.67% coverage. All 120 Lexe contract cases passed. Full root suite: 11098 passed, 471 skipped, 11 xfailed. Root and funding lint/type checks passed. All required CI gates passed.

## Release
Built committed source d1b292de with the existing tr-deploy identity; no IAM changes. Image digest sha256:1562d3a05ba6667c99a7e236699a8b2d20e6b70aa99a2bdcd4037e5af80ee9f7. Staged revision lightning-router-web-clockd1b292de is Active, Ready, ContainerHealthy and reports healthy receive authority. Configuration compared exactly to the prior revision except image. Production promotion and recovery verification are in progress. Alerts remain strict.


jperla:
Production verification complete: fixed revision lightning-router-web-clockd1b292de serves 100% of traffic. The normal worker reconciled the original flagged invoice to CANCELED, cleared its failure code, and left zero credit, no deposit, and no customer account. This was verified with one read-only primary-key Cloud SQL query. Public /health is HTTP 200 with payments_ready=true, review_required=0, uncredited_count=0. Homepage, usage, pricing, and docs all return 200. Wallet/secret/network/runtime configuration is unchanged; no manual ledger edits, outgoing payments, IAM grants, or alert threshold changes were made. All CI gates passed and the full local suite passed (11098 tests); funding passed 688 tests across SQLite/PostgreSQL.

PR #1239, opened 2026-09-20
Gateway: racing idempotent replay on the legacy path leaks the loser's key hold

jperla:
> **MERGE GATE — do not merge yet.** The orphan-credit reclaim is exclusive only once no worker that predates #1243 is running: a previous-version worker hands reservations out by the bare idempotency key, so during a mixed rollout it could attach a reservation that this change then reclaims. Merge and deploy this only after #1243 is verified live on **both** Postgres planes (AWS and Azure) with previous-version tasks drained. `main` is what those planes deploy, so merging earlier would ship both changes in one rollout.

## Bug (pre-existing, found while reviewing #1120/#1122)
On the **non-typed** authorize path (memory and Postgres stores; Postgres is the AWS/Azure backend), two requests with the same `Idempotency-Key` can both pass the pre-probe. Each escrows a key hold with `STORE.reserve_key_limit` (committed outside any transaction), and the second `create_gateway_authorization` returns the **first** one's row and writes nothing. The loser's key hold was never released: the reaper only sees authorizations.

The typed Spanner path is **not** affected and is untouched.

## Fix (fourth version, on top of the reservation-scope change #1243)
The credit reservation is keyed by idempotency key and therefore **shared**: the loser's `reserve()` returns the winner's reservation and places no second hold. So what leaks is the loser's **key** hold, plus, in one case below, an orphaned credit reservation.

When the store returns an authorization whose id is not the one this request minted:

- release the user-model slot;
- refund **this attempt's key hold**, and only when one was actually taken (`reserved_microdollars > 0`), for the recorded amount rather than the estimate;
- reclaim a credit reservation **only** when it is not the one the winner's authorization references **and** the `Reservation` returned by `reserve()` is owned by this workspace and API key. This covers a `deferred_home` winner that holds no local reservation while the loser, after a credit transfer landed, reserved locally. The winner's reservation is never touched;
- `409` on fingerprint mismatch, otherwise the stored replay response with `idempotent_replay: true`.

Each refund is **one call**. The stores already own the conflict retry budget (Postgres retries serialization conflicts internally before raising `StoreConflict`), so an outer retry only multiplied it on a request thread. A failure is logged once as `billing.replay_key_refund_failed` / `billing.replay_credit_refund_failed` with `outcome=rolled_back` (`StoreConflict`: the store proved the transaction did not commit) or `outcome=unknown`, and is never retried or re-raised: after an unknown commit outcome a blind retry could refund twice. A refund failure never changes the replay response or the 409. Durable recovery of an unknown-outcome refund needs attempt-scoped key holds in the stores; this change does not add them and says so in the code.

**Why the reclaim is exclusive.** With #1243 the reservation index has the same scope as the authorization index, so every attempt that can receive a reservation maps to one workspace / API key / idempotency slot. In the replay branch the winner already owns that slot, idempotency pointers are never deleted, and a replay writes nothing, so no later attempt can attach an unreferenced reservation.

## Why the earlier versions were withdrawn
**v1** gave every attempt its own credit reservation. Independent review reproduced four regressions against `main`: a `StoreConflict` during authorization creation left one estimate reserved forever (main heals it because the retry adopts the keyed reservation); pause tombstones skipped cleanup of the second private reservation; two cleanup calls could fail halfway; and a matching replay needed credit for a second reservation, so it returned 402 where main returns 200.

**v3** (blocked): on the old bare-key reservation index a request under a *different* API key with the same `Idempotency-Key` could be handed the loser's reservation, and the reclaim then unfunded it. That index is fixed by #1243, which this version is rebased on; the scenario is now a test. v3 also retried refunds in an outer loop, multiplying the store's own retry budget.

**v2** kept the shared reservation and refunded only the key hold. Review found the deferred-winner orphan above and that a swallowed refund failure was not even retried when the store had provably rolled back.

## Tests
`tests/test_legacy_authorize_replay_leak.py`, **118 tests on memory and Postgres**, at two interleavings for the ordinary race and two for the deferred winner. One authorization; both counters asserted through one helper; winner's credit reservation live and unsettled; the replay response carries the **winner's stored** authorization id, estimate, model, provider and endpoint even when the loser computed different values; zero residue after settlement; 409 on mismatch with only the winner's holds left; deferred winner plus local loser leaves no credit hold; a second API key and a second workspace sending the same idempotency key before, during and after the reclaim always get their own reservation and stay funded; a reservation returned from an earlier failed create is reclaimed exactly once; uncapped key, cap removed between attempts, recorded-hold-not-estimate; one call per refund on success, on `StoreConflict` and on an arbitrary exception, with the log line asserted.

Mutation check, each reverted separately, all caught: key refund removed (102 tests fail); orphan reclaim removed (42); winner-reference guard removed (58); ownership guard removed (8); positive-hold guard removed (4); refund amount changed to the estimate (4); refund exception re-raised (24); response rebuilt from the loser's request (54); reclaim fires on the winner's reservation (58).

Full suite as CI runs it, on a285d155: **11,864 passed, 0 failed**. ruff and mypy clean.

## Known, tracked separately
The stores' pause-rejection path releases the estimate rather than the attempt's recorded key hold, and a pause-tombstone replay inside `create_gateway_authorization` returns without releasing the attempt's key hold. Both pre-date this change, live in the stores rather than the gateway, and get their own PR.

## Review status
v1, v2 and v3 were each blocked or sent back by independent review (findings above). Review of v4 (70adb1be): all three earlier findings confirmed **fixed** on memory and real PostgreSQL (the cross-key reclaim no longer reproduces for fully upgraded workers; 8 transaction attempts per refund instead of 24; the stored-response test now fails when the response is rebuilt). The one remaining blocker is the rollout-ordering gate at the top of this description, which needs no code change. aa60f090 only adds that precondition to the code comment.


jperla:
Holding this. Independent review reproduced that the unkeyed credit reservation introduced here leaks credit in cases `main` heals through same-key retry (StoreConflict during authorization creation, pause tombstones, a replay that now needs credit for a second reservation and returns 402 instead of 200). Reworking to the smaller fix: keep the shared keyed reservation, and on a detected replay refund only the loser's **key** hold, which is the only thing that actually leaks on `main`. Regression tests for each reviewer scenario are being added. Do not merge until the re-review is clean.

PR #1240, opened 2026-09-20
Fix premature stream probe timeouts and catalog-dependent billing tests

jperla:
Follow-up to #1237. The rotation probe incorrectly applied its first-token timeout to the entire stream. It now switches exactly once after actual content/reasoning arrives to the existing absolute completion deadline. Heartbeats cannot extend the first-token budget; repeated tokens cannot extend the completion budget. Cancellation still propagates and closes the stream. No global timeout or alert threshold increased.

Price-refresh run 35481412597 exposed three exact-operation tests depending on the optional GMI Haiku BYOK endpoint. Removing that endpoint reproduced all three failures locally. Fixed prepaid/BYOK fixtures retain exact operation-count assertions without changing billing code.

Regression evidence: four new stream cases fail on old code, then pass after repair; focused suite passed (155 tests before the additional cancellation test). Ruff and mypy (382 source files) pass. Full suite running. Public checker companion: Lore-Hex/trustedrouter-provider-check#7, 259 tests pass. Live NVIDIA recovery is not yet verified and its alert stays open. No refunds, disputes, IAM, or attestation policy changes.


jperla:
Local full-suite verification completed: 11,167 passed, 472 skipped, 11 xfailed; Ruff clean; mypy clean across 382 source files. Public checker companion #7 merged after Python 3.11/3.12/3.13 CI and wheel gates passed, with 259 local tests. Cross-repository contract parity passes. Initial local attempt was invalidated by deletion of the shared pytest temp directory and denied loopback sockets; the complete rerun used an isolated temp directory and local mock-server access. No guard was weakened.

PR #1241, opened 2026-09-20
Reduce snapshot worker upload by 82% without weakening transfer gates

jperla:
Deployment 35481392420 hit its 300-second SCP limit. Read-only inspection confirmed successful SSH authentication, only two project SSH keys, and a truncated 14,100,480-byte archive. The live previous snapshot worker continues publishing five snapshots every minute.

Use the existing committed-source/UTF-8 JSON validated bundle helper for public snapshot sync, excluding only website static assets. Equivalent bundles are 14,484,681 versus 2,581,099 bytes. Isolated extraction/import successfully loaded the worker and all 625 catalog models. Full worker bundles retain existing contents. Unknown bundle scopes fail closed.

Unchanged: 300-second upload bound, no mutation after failed transfer, ten-minute CI SSH keys with cleanup, remote rollback and five-snapshot publication checks. No IAM or alert threshold changes.

Validation: 57 focused deployment tests; Ruff and mypy clean; shell syntax clean. Full local suite running. Production transfer remains to be verified after protected rollout.


jperla:
Full local suite passed: 11,169 tests, 472 skipped, 11 xfailed. Ruff and mypy (382 files) pass. Shell syntax passes. The reduced committed bundle was extracted into an isolated directory and the actual snapshot-worker imports and 625-model catalog loaded successfully without static assets.

PR #1244, opened 2026-09-20
Accelerate snapshot IAP uploads and expose failed publication

jperla:
## Summary
- Run snapshot publication inside a pinned, isolated NumPy-enabled Python so gcloud uses the supported IAP upload accelerator.
- Report raw snapshot-sync failure in workflow warnings and the release summary instead of mistaking continue-on-error for successful publication.
- Preserve the 300-second transfer limit, committed-source validation, short-lived SSH access, rollback, and five-product verification. No IAM, payment, or serving runtime changes.

## Evidence
The prior upload timed out after transferring 2,350,080 of 2,580,921 bytes. Node CPU, memory and disk were healthy. Smaller archives alone did not fix it. This is a targeted mitigation pending real production upload verification, not a claim that network root cause is conclusively proven.

## Tests
New regressions failed before the change. Workflow and bundle tests pass; Ruff and mypy pass. Isolated gcloud runtime validated locally. Full suite is running; results will be added before merge.


jperla:
Local gates complete: 11,171 passed, 472 skipped, 11 xfailed; Ruff and mypy (382 files) pass. The first local attempt was blocked by sandbox loopback restrictions; the complete rerun with fake-server loopback access passed. The pinned isolated Python runs gcloud successfully, and a 2,580,921-byte zero-data diagnostic over the accelerated IAP path reached a byte counter in 1.74s without files or database changes. This is not yet a successful CI worker publication; that remains the production acceptance check.

jperla:
The same-size control transfer using unaccelerated gcloud also completed quickly (1.48s). This rules out node capacity and basic IAP reachability at this moment, but does NOT establish missing NumPy as the original root cause. Keep this labeled as upload-path hardening until the actual GitHub-runner SCP and worker publication succeed. Neither transfer modified node files or analytics.

PR #1250, opened 2026-09-20
decide: three fast names, /api/alpha/decide as the route's name, Gemma becomes gemmev

jperla:
## What

Follow-up to #1235 and #1249. Three things Joseph asked for.

### 1. Fast names

Low latency is most of why anyone wants a decision model, and only `trev-1.0` and Jev were under half a second. Each of these was measured four times on the live eval, every output valid:

| Name | Backed by | Hosts, in order | Median | Eval | About one decision |
|---|---|---|---|---|---|
| `trustedrouter/mev-1.0` | `inception/mercury-2`, reasoning off | inception | 275-310 ms | 28/29 (29/29 at `low` effort, ~600 ms) | $0.00017 |
| `trustedrouter/lev-1.0` | `meta-llama/llama-3.3-70b-instruct` | sambanova → parasail → together | 371-450 ms, worst case 565 | 29/29 | $0.00036 |
| `trustedrouter/zev-1.0` | `z-ai/glm-5.2-fast` | fireworks → baseten | 0.35-0.7 s | 29/29 | $0.0014 |

They are rows in `NAMED_DECISION_MODELS`, so they get everything trev has because it is the same code. Their plain ids join the tuned native lists.

Measured and left out: GLM 5.3 Flash cannot turn thinking off (2-19 s on every host), and Qwen 3.8 27B on Cerebras was the fastest native (216-301 ms) but has no usable fallback behind a rate-limited host. Mercury 2 had earlier been rejected for truncated output; that was a config mistake (it reasons by default), not the model.

### 2. A rename

`mev-1.0` meant Gemma 4 E4B when it shipped a few hours ago. Joseph gave the name to Mercury 2; Gemma is now `trustedrouter/gemmev-1.0`. The test that pins the promised names now pins every (id, backing model) pair and every multi-host chain, and says why this must not be done to a name with traffic.

**Anyone who hardcoded `mev-1.0` in those few hours now gets Mercury: faster, and about 8x the price per decision ($0.00017 vs $0.00002).**

### 3. The route's name is `POST /api/alpha/decide`

"alpha" because the contract may still change. It moved once already and a page kept advertising the old path, so it now has one source: `DECIDE_PATH`, `DECIDE_PATH_ALIASES` and `decide_url()` feed the docs, the shared curl example, `/docs`, authorize's refusal message, the comparison FAQ and `llms.txt`; a test fails if any template spells either path by hand.

Aliases the gateway answers to, all listed in the docs: `/api/decide`, `/v1/decide`, `/v1/evaluate` (Vercel), `/api/alpha/decisions` and `/api/decisions` (OpenRouter's Decisions API, which now serves Jev). Callers coming from OpenRouter or TypeSafe can send `"type": "noul"` and get `{"type":"noul","noul":p}` back.

## Testing

- Full suite: 11,530 passed (472 skipped and 11 xfailed are pre-existing). `ruff check` and `mypy` clean.
- Every named-model test is parametrized over all eight names.
- New tests were checked by removing what they guard (5 mutants, all killed).
- Live eval of the whole shipped table against the real hosts (gateway repo): all 15 entries 8/8 valid; 29/29 everywhere except mev's 28/29.
- codex review: PASS on round 2 (round 1 found `/docs` still showing the old path).

## Rollout

The gateway half (Lore-Hex/quill-cloud-proxy `1a03c16`, codex PASS) was pushed first and takes about two hours. I am holding this merge so the docs do not advertise `/api/alpha/decide` before every gateway region serves it. While the two sides disagree about `mev-1.0`, a request for it is refused cleanly (the hosts asked for and the enforced chain do not intersect); it never runs on the wrong model.


jperla:
Added one commit after calling production: a caller's `provider` filter that excludes every host of a name now gets an immediate 400 instead of the chain's 503 (which told the caller to retry and cost 7 s of gateway retries). codex PASS; full suite 11,538 passed. Merge is still held until the gateway deploy (blocked on a GCP zone stockout) lands.

jperla:
Rebased onto main. My extra commit (the 400 for a provider filter that excludes a name's hosts) is dropped: #1257 landed the same rule on main first. This PR is back to the one reviewed commit; full suite 11,582 passed on the rebased branch.

PR #1253, opened 2026-09-20
Put Token Exchange cities above the hero

jperla:
## Summary
- Show all eight city exchanges directly below the masthead on all twelve market pages.
- Keep global/region links secondary, highlight the current market, and retain the full footer directory.
- Add navigation placement, completeness, current-state and overlap/hit-target coverage.
- Document independent global/New York HTTPS certificates.

## Verification
- Regression test failed on the previous template for all twelve markets before implementation.
- Nine static-site unit tests pass.
- All 36 desktop/tablet/mobile browser cases pass; first-viewport screenshots visually inspected.
- Ruff and mypy pass. Full application suite is covered by CI; redundant local run stopped because the sparse, shared environment is not the CI environment.
- Public HTTPS smoke passes for all 29 domains and www aliases, all 12 market pages, robots, sitemaps and OG assets, plus existing homepage/trust/status.
- All regional certificates and the independent New York certificate are ACTIVE.

## Release
Static GCS/CDN content publish only after CI passes. Existing DNS, redirects, inference paths and billing are unchanged.


jperla:
TLS recovery verified: all four regional batches and both independent flagship certificates are now ACTIVE. Public HTTPS smoke passed all 29 domains plus www aliases, all 12 canonical pages, robots/sitemaps/OG assets, and the existing TrustedRouter homepage, trust and status surfaces. The one failure seen in the redundant local full-suite run was a nested uv dependency-sync attempt in the restricted shared environment; rerunning that test with UV_NO_SYNC=1 passes. No application code changed.

jperla:
Deployed b2f828d as a static content-only publish to gs://quill-token-exchange-public. Refreshed only the twelve exchange homepage CDN entries. Verified normal public roots contain all eight city links above the hero on all twelve markets. Production browser checks passed at 390/768/1440px for the global and New York sites, with screenshots visually inspected, valid TLS, loaded artwork and no horizontal overflow. Public smoke verifies canonical pages, robots, sitemaps, OG assets and all domain aliases; existing TrustedRouter/trust/status remain healthy.

PR #1256, opened 2026-09-20
Advertise 100+ providers on the homepage

jperla:
## Summary
- Change the homepage provider statistic from 90+ to 100+.
- Update the two server-rendered homepage assertions and browser assertion.
- No layout, routing, billing, catalog, or unrelated copy changes.

## Verification
- Updated homepage assertions failed on the old template and pass with the change.
- Homepage browser test passes, including sign-in flow and existing layout checks.
- Ruff and mypy pass locally.
- Full local pytest suite is running; CI must pass before merge and normal production deployment.

## Release
Use the standard gated control-plane workflow. Verify the public homepage shows 100+ providers and existing health/status/exchange surfaces remain healthy after rollout.


jperla:
Local verification: ruff and mypy pass; both updated homepage tests first failed on the old template and now pass; the homepage Playwright test passes. The full local suite was attempted but stopped after widespread setup errors with the laptop down to about 253 MiB free. It is not reported as a passing full local run. This four-line template/test-only change still requires the complete clean-runner CI and normal deployment gates.

jperla:
Production verified. The initial release was superseded by a newer main commit that includes this change; the replacement deployment completed successfully: https://github.com/Lore-Hex/quill-router/actions/runs/35532262230 . The normal public homepage now renders 100+ providers. Live Chromium checks pass at 390px and 1440px with screenshots visually inspected and no horizontal overflow. Homepage health, trust, status, New York Token Exchange and the global Token Exchange all return 200. All CI/release gates passed; no gates were bypassed.

PR #1258, opened 2026-09-21
test: monitor OpenCode chat store=false compatibility

jperla:
## Release order: HOLD until gateway rollout is verified
Depends on Lore-Hex/quill-cloud-proxy#351 (gateway commit 04b5665298e473d1aca858b6c61e35e0c5ebbf25). Do not merge while any monitored gateway still rejects store=false. Gateway release run: https://github.com/Lore-Hex/quill-cloud-proxy/actions/runs/35590791655 . GCP rollout previously encountered TDX capacity constraints; never bypass attestation or deploy gates.

## Summary
- Add store=false to the existing SDK Chat PONG request, matching OpenCode / @ai-sdk/openai-compatible.
- No extra inference calls, tokens, or cadence changes.
- Exercise 400 router errors and prove the existing repeated-failure Sentry path fires at its bounded threshold, then remains quiet. Keep key and private error detail out of samples/alerts.
- Do not send arbitrary user 400s directly to Sentry or add a Sentry SDK to the enclave.

## Verification
- Exact request-shape regression failed before the probe change.
- 49 targeted SDK/alert tests pass; full-suite run is in progress.
- Full ruff and mypy gates pass.

## Root cause
The enclave has metadata-only request_contract_rejected logging but no Sentry event on that path. Existing PONG probes omitted store, so customer SDK compatibility failures did not trigger their debounced Sentry alarms.


jperla:
Validation completed: full local pytest suite passed (11,752 passed, 472 skipped, 11 xfailed); full ruff and mypy passed. The request-shape regression was also verified red before the change, and all 49 focused SDK/alert tests pass.

Release hold remains: gateway fix #351 in Lore-Hex/quill-cloud-proxy is merged as 04b5665298e473d1aca858b6c61e35e0c5ebbf25. Its normal release 35590791655 built the image and passed the signed public transition-policy gate, but the US Central update has a replacement in CREATING with ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS. Old instances remain running. Do not merge this monitor change until the fix is verified on all monitored chat targets; do not bypass capacity, attestation, or rollout safety gates.

